Issue #1, opened 2019-08-09
Radio button bullet don't rendered properly


edimitchel:
Hello,
first, I really like your components set ! I hope you will provide more components and more mixins.

And now, I saw an issue on Radio buttons. Bullets are not rendered properly on my computer : Windows 10 with **Firefox**.

![image](https://user-images.githubusercontent.com/2922851/62765777-237e1300-ba91-11e9-8dc0-95a5be6a9f89.png)

Thanks !

kiaking:
Hi @edimitchel, thank you so much for the report!

This is clearly not good. I've checked with Firefox on Mac, and it seems OK so I guess it's Windows specific one. I'll look into it and fix it 👍

edimitchel:
I can propose you a fix, it could be simpler for testing !

kiaking:
@edimitchel Hi! Sorry for my late response. It was a big holiday week in Japan.

Yes! That would be super awesome if you can fix it!

kiaking:
I tried to reproduce this with Windows 10 Firefox on VirtualBox, but it didn't. Do you have any other specific idea why this might be happening?

edimitchel:
It seems my screen make the bullet in normal zoom. But when I zoomed out, the bullet is centered.

#### Unzoomed
![image](https://user-images.githubusercontent.com/2922851/63672316-fbf1af00-c7e1-11e9-8d2c-2bc67bb2101d.png)

#### Zoomed (80%)
![image](https://user-images.githubusercontent.com/2922851/63672338-08760780-c7e2-11e9-9987-ac0b15f992bc.png)

Just for information, I have the ESR 60 Firefox version.

kiaking:
Thank you for the hint! In my case, the dot became strange on zoom 90%, but both 100% and 80% looks OK. I guess border-radius 50% and position absolute is doing some bad things.

I'll try to tackle this one to see if I can fix on my browser's different zoom levels.

kiaking:
@edimitchel Hi! I've created a PR #2 for the fix. Could you try it out on your environment? You can clone the repo and run `yarn serve` to test it out.

kiaking:
@edimitchel OK, I've updated the #2 to use `transform: scale` instead of `position: absolute` approach. Could you check with the latest #2 and see if it's fixed?

kiaking:
This issue is fixed by Sefirot 0.6.2! Thanks again @edimitchel for the report! Please let us know if there's anything 👍 
[...]

PR #2, opened 2019-08-27
#1 Fix input radio circle align center at different zoom level

kiaking:
Issue #1

This PR fixes InputRadio circle to be aligned at center in different zoom level.

### DIFF
--- a/lib/components/inputs/SInputRadio.vue
+++ b/lib/components/inputs/SInputRadio.vue
@@ -9,7 +9,9 @@
     <div class="container">
       <div class="input" :class="{ on: value }" role="button" @click="emitChange">
         <div class="box">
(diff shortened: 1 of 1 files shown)


edimitchel:
Hi ! I test it out, but it still has rendering issue with my firefox.. I don't have the latest version of Firefox, I have the 60th (the latest of ESR kind).

kiaking:
@edimitchel Hi! Thank you for the confirmation. Hmm... Let me see, maybe I came up with another approach using `transform: scale` instead of absolute positioning. I'll push the fix soon and let you know again!

kiaking:
Great! Thank you for checking. I'll merge and make a new release 👍

PR #49, opened 2021-03-18
fix: add missing placeholer props to SInputDate

ryo-gk:
Add missing props `placeholder` to `SInputDate`.


kiaking (approved):
Uh nice! thanks 👍

PR #51, opened 2021-03-25
feat: add 'step' props to input number

ryo-gk:
## Proposal
Add `step` props to `SInputNumber` .
This is usable to input a decimal value.

If input decimal value without decimal step attribute,
waring message is shown.
image


kiaking (approved):
Wo this is cool. I didn't know about `step` attribute. Nice 👍

PR #52, opened 2021-03-26
feat(validation): include() can receive array as 'other' argument

ryo-gk:
## Proposal
Allow `include` validation to receive array as `other`.


ryo-gk:
I created other PR instead of this one. 
https://github.com/globalbrain/sefirot/pull/53

PR #53, opened 2021-03-26
feat(validation): implement includeSome

ryo-gk:
## Proposal
Implement `includeSome` validator which is like `include`.
The difference is this validator can receive array as `other` argument,
and return true if `value` includes one or more of given another value.


kiaking (approved):
Nice!

PR #56, opened 2021-04-09
feat: narrow form validation type

cuebit:
Intersects the form validation property as a required key when using rules.

useForm

### DIFF
--- a/lib/composables/Form.ts
+++ b/lib/composables/Form.ts
@@ -21,7 +21,7 @@ export type UpdateFunction<T extends Record<string, any> = Record<string, any>> 
   value: any
 ) => void
 
-export type FormValidation<T, U extends Form<T> = Form<T>> = WithRequired<U, 'validation'>
+export type FormWithValidation<T, U extends Form<T> = Form<T>> = WithRequired<U, 'validation'>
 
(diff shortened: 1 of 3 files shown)


lib/composables/Form.ts
```diff
@@ -10,18 +11,33 @@ export interface Form<T extends Record<string, any>> {
...
   rules?: Rules
 }
 
-export function useForm<T extends Record<string, any>>(options: UseFormOptions<T>): Form<T> {
+export type UpdateFunction<T extends Record<string, any> = Record<string, any>> = (
+  model: keyof Data<T>,
+  value: any
+) => void
+
+export type FormValidation<T, U extends Form<T> = Form<T>> = WithRequired<U, 'validation'>
```
kiaking, line 24: Maybe this should be `FormWithValidation`? `FormValidation` felt a bit more like it's only validation, without `data`. 👀

kiaking (commented):
Niiiice! Added a comment!

lib/composables/Form.ts
cuebit: I could of sworn it was 😆 ... I'll revert!

kiaking (approved):
👍 👍 👍

PR #57, opened 2021-04-11
feat: markdown component

cuebit:
## Overview

This PR introduces a standalone component to support markdown rendering. The component requires and implements [markdown-it](https://github.com/markdown-it/markdown-it) as it's choice of renderer. The component can be immersed within existing components as well as externals. The composition is designed to work flexibly to allow custom implementations if necessary.

Basic example:
```vue
<SMarkdown content="Hello **world**" />

<div class="SMarkdown-container">
  <p>Hello <strong>world</strong></p>
</div>
```

## Props

| Prop | Type | Default | Description
| --- | :---: | :---: | --- |
| `content` | String | – | **Required** - the raw markdown string. 
| `inline` | Boolean | `false` | Renders content without wrapping in `<p>` tags.
| `tag` | String | `'div'` | Renders a custom root element. Useful when combined with `inline`.
| `callbacks` | Array | `[]` | A list of callbacks for custom links. Callbacks are referenced in markdown links by their index within curly braces. References can be repeated as many times e.g. `[link1]({0}) [link2]({1}) [link3]({0})`. Undefined or out-of-range references will throw.

## Slots

By way of simplicity, the component will render markdown using the `v-html` directive. This means component children will be replaced by the rendered HTML. Subsequently, the component exposes some slot props to help customize the component and its dependents. This allows for greater flexibility for nested components.

```vue
<SMarkdown v-slot="{ rendered, selector }" v-bind="{ content, inline, callbacks }">
  <SCard>
    <SIconCode />
    <p :class="selector" v-html="rendered" />
  </SCard>
</SMarkdown>
```

### Slot Props

| Prop | Type | Description
| --- | :---: | --- |
| `selector` | String | The class name that should be applied to a wrapping element of the rendered html. This is necessary to prevent event listener pollution in the DOM and avoids any side-effects where listeners may hijack anchors they are not supposed to.
| `rendered` | String | The rendered markdown markup. To be used with the `v-html` directive.

#### `clicked`

Emitted when a link is clicked. Allows for arbitrary code execution in components to extend functionality.

```vue
<SMarkdown content="[link](/route)" @clicked="(payload) => console.log(payload)" />
```

See the `useLink` subscribe example below for info on payload types.

## Composition

There are two composition functions exposing APIs which are used by the component and can also be freely used to compose custom components. In addition, the compositions possess additional API functionality for more bespoke handling.

### `useMarkdown`

Options: `{ linkAttrs?: LinkAttrs; config?: (md: MarkdownIt) => void }`
Returns: `(source: string, inline: boolean) => string`

Instantiate and (optionally) configure the markdown renderer. The `markdown-it` instance can be manipulated further by passing a config callback to the initiator:

```ts
// configure renderer
useMarkdown({
  config: (md) => {
    md.use(somePlugin)
      .set({ breaks: true })
  }
})
```

Returns a factory method that, when invoked, will render the given source (markdown string). The second argument, default `false`, renders the source as inline.

### `useLink`

Options: `{ callbacks?: LinkEvent[] }`
Returns: `{ selector: string; addListeners: Function; removeListeners: Function; subscribe: Function }`

Exposes an API to aid event listening for rendered links. The internal handler does nothing if 1) there is no `href` present in a link, and 2) if it's an external link. If a callback link is triggered, it will invoke the respective callback, and will throw if one is not found. Otherwise, the router will handle the result.

The `selector` is a computed string with a UUID suffix to distinguish between multiple DOM elements (e.g. `SMarkdown-ah42wenv`).

The `addListeners` and `removeListeners` bind/unbind the necessary event listeners to links within containment. Listeners are removed automatically when a component is destroyed but the method is exposed for flexibility.

The `subscribe` method allows for arbitrary subscription to events. All subscriptions receive a payload and are executed before the internal handler decides what to do. This offers code execution that is decoupled from the standard handler behaviour.

```ts
const { subscribe } = useLink({ ... })

subscribe(({ event, target, isExternal, isCallback }) => {
  console.log(event.type) // 'click'
  console.log(target)     // HTMLAnchorElement
  console.log(isExternal) // boolean
  console.log(isCallback) // boolean
})
```

## Progress
- [x] Component
- [x] Composition
- [x] Spec
- [ ] Documentation

### DIFF
--- a/lib/components/SMarkdown.vue
+++ b/lib/components/SMarkdown.vue
@@ -1,0 +1,38 @@
+<template>
+  <component :is="tag" v-if="$scopedSlots.default" class="SMarkdown-container">
+    <slot v-bind="{ rendered, selector }" />
+  </component>
+  <component :is="tag" v-else :class="['SMarkdown-container', selector]" v-html="rendered" />
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, nextTick, PropType, watch } from '@vue/composition-api'
+import { useMarkdown, useLink, LinkEvent } from '../composables/Markdown'
+
+export default defineComponent({
+  props: {
+    content: { type: String, required: true },
+    callbacks: { type: Array as PropType<LinkEvent[]>, default: () => [] },
+    inline: { type: Boolean, default: false },
+    tag: { type: String, default: 'div' }
+  },
+
+  setup(props) {
+    const markdown = useMarkdown()
+    const { selector, addListeners } = useLink({ callbacks: props.callbacks })
+
+    const rendered = computed(() => markdown(props.content, props.inline))
+
+    watch(
+      rendered,
+      () => nextTick(() => addListeners()),
+      { immediate: true }
+    )
+
+    return {
+      selector,
+      rendered
+    }
+  }
+})
+</script>
--- a/lib/components/SMarkdown.vue
+++ b/lib/components/SMarkdown.vue
@@ -17,9 +17,9 @@ export default defineComponent({
     tag: { type: String, default: 'div' }
   },
 
-  setup(props) {
+  setup(props, { emit }) {
     const markdown = useMarkdown()
-    const { selector, addListeners } = useLink({ callbacks: props.callbacks })
+    const { selector, addListeners, subscribe } = useLink({ callbacks: props.callbacks })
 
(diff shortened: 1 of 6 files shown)


kiaking (approved):
This is looking really good!!! 🚀 🚀 🚀 🚀 🚀

PR #60, opened 2021-04-16
refactor(markdown): replace `selector` with container `ref`

### LINKED ISSUE #59: refactor(markdown): replace `selector` with container `ref`
The `SMarkdown` component should use a `ref` to reference the root node and apply event handlers using said reference.

Currently the `useLink` composable generates a unique selector to allow slots to explicitly apply the class to a target container for better listener binding. But the internals explicitly target `.SMarkdown-link` making the selector an technical burden and completely redundant.

cuebit:
resolves #59


kiaking (approved):
LGTM!

PR #65, opened 2021-04-28
feat: add tabindex to input-selent and input-dropdown

ryo-gk:
Now `SInputSelect` and `SInputDropdown` can't be focused with tab key.
So enable to focus with tab key by attaching `tabindex`.

### Select
**mode:filled**
Screen Shot 2021-04-28 at 20 56 16
**mode:outlined**
Screen Shot 2021-04-28 at 20 54 10

### Dropdown
**mode:filled**
Screen Shot 2021-04-28 at 21 24 54
**mode:outlined**
Screen Shot 2021-04-28 at 21 23 28

### DIFF
--- a/lib/components/SDropdown.vue
+++ b/lib/components/SDropdown.vue
@@ -14,13 +14,27 @@
         :placeholder="options.search.placeholder"
         :value="options.search.value.value"
         @input="v => options.search.onInput(v)"
+        @down="focusFirstItem"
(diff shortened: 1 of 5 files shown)


kiaking:
Thanks for kicking start on this! I've updated to also add keyboard control. For example, now we may able to navigate the dropdown items by up/down key 👍

PR #67, opened 2021-04-30
feat: explicit modal toggling

### LINKED ISSUE #66: Implement explicit modal tracking
Currently, modal actions `push` new modals to the state when opened and `pop` modals from the state when closed.

```ts
store.dispatch('modal/open', payload)
store.dispatch('modal/close')
```

The modal API may be improved by an optional `uid` prop assignment. When used in composition, modals can be assigned an auto-incrementing `uid: number` for simplicity therefore can be "dropped" from the state regardless of its position in the stack.

```ts
let nextId = 0

function useModal() {
  const store = useStore()
  const uid = ++nextId

  function open(payload): void {
    store.dispatch('modal/open', { ...payload, uid })
  }

  function close(): void {
    store.dispatch('modal/close', uid)
  }

  return { open, close }
}
```

cuebit:
resolves #66

Breaking changes for `useAlert` which no longer requires an argument on invocation.

### DIFF
--- a/lib/components/SModal.vue
+++ b/lib/components/SModal.vue
@@ -26,7 +26,8 @@ export default defineComponent({
     data: { type: Object, default: () => ({}) },
     show: { type: Boolean, required: true },
     width: { type: String, default: 'auto' },
-    closable: { type: Boolean, default: true }
+    closable: { type: Boolean, default: true },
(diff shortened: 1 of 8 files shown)


kiaking (approved):
Niiiice! Adjusted few stylings. Very nice 👍

PR #71, opened 2021-05-11
fix(input-text): clearable & disabled should not conflict

### LINKED ISSUE #70: Disabled inputs show interactive clearable icon
Inputs with `clearable` and `disabled` enabled should not have the clear button visible which leads to emitting.

cuebit:
Disabling emitting and hiding the button seems overkill. Hiding the button logic already exists, so pointer-events can be disabled to prevent emits.

resolves #70


kiaking (approved):
Looking nice! Thanks a lot 🙌

PR #75, opened 2021-05-21
fix(modal): handle active state timer (#74)

### LINKED ISSUE #74: Modal active state incorrectly set
Modal items pushed to the stack _after_ the final item is closed are not accounted for when the component is lazily marked as inactive.

This is [due to a setTimeout](https://github.com/globalbrain/sefirot/blob/438699e0bafff86fd596a3fab1e942a072920a8c/lib/components/SPortalModals.vue#L42) marking the `isActive` flag as `false` therefore any modals pushed to the stack before the timer expires are ignored and do not appear as active.

The timer should be cleared before it is created to avoid making the assumption the stack is foreseeably empty.

cuebit:
resolves #74 

Additionally, modal compositions return store actions for awaitable procedures.

### DIFF
--- a/lib/components/SPortalModals.vue
+++ b/lib/components/SPortalModals.vue
@@ -76,12 +76,12 @@ export default defineComponent({
   bottom: 0;
   left: 0;
   z-index: var(--z-index-backdrop);
-  background-color: rgba(0, 0, 0, 0.8);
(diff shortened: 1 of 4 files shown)


kiaking:
Awesome! Thanks!

PR #77, opened 2021-05-25
feat(card-header): add disabled option for header action

ryo-gk:
Add `disabled` property for action of Card component.

The usage is like this.

```js
{
  header: {
    title: 'Some title',

    actions: [
      {
        type: 'text',
        mode: 'info',
        icon: SIconEdit3,
        disabled: 'Cannot click. It is disabled.',
        callback: () => {}
      }
    ]
  }
}
```

`disabled` property accepts string, and it will be shown as a tooltip when an icon is hovered.

Once the icons for `disabled` are ready, we can mark this PR as ready.


lib/components/SCardHeader.vue
```diff
@@ -54,8 +76,22 @@ export default defineComponent({
       `round-${props.round}`
     ])
 
+    function getIcon(icon: any, disabled: boolean) {
```
kiaking, line 79: Let's import `ActionIconType` from `Card` and use it here 🙌 

```suggestion
    function getIcon(icon: ActionIconType | object, disabled: boolean) {
```

lib/composables/Card.ts
```diff
@@ -22,12 +22,15 @@ export interface Footer {
...
   mode?: Mode
-  icon?: any
+  icon?: ActionIconType | object
+  disabled?: string | null
   label?: string
   link?: string
   callback?(): void
 }
 
+export type ActionIconType = 'add' | 'edit' | 'delete'
```
kiaking, line 32: Let's change the option name to match the icon file 🙏 

```suggestion
export type ActionIconType = 'plus' | 'edit-3' | 'trash-2'
```

Issue #79, opened 2021-06-10
[InputYear] handle null value


ryo-gk:
When the user inputs an empty value, be emitted `Number(null)` that is Nan.
This cause unexpected behaviors in `year()` validation.

PR #80, opened 2021-06-10
fix: emit number only when value is not NaN (#79) 

ryo-gk:
close https://github.com/globalbrain/sefirot/issues/79

Emit number value only when the value can be converted to a number that is not NaN.


kiaking:
Let's use `SInputNumber` instead so that we don't have to worry about parsing `NaN`. I think we didn't have `SInputNumber` when I created this component 😅

PR #84, opened 2021-06-18
fix: make button unclickable if disabled

ryo-gk:
Make the disabled button not clickable.

### DIFF
--- a/lib/components/SCardHeader.vue
+++ b/lib/components/SCardHeader.vue
@@ -154,6 +154,10 @@ export default defineComponent({
 .action.warning:hover { color: var(--c-warning); }
 .action.danger:hover  { color: var(--c-danger); }
 
+.action.action.disabled {
+  pointer-events: none;
+}
+
 .action.disabled:hover {
   color: var(--c-text-2);
   background-color: transparent;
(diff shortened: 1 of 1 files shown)


kiaking (commented):
I think CSS style is good add as well, but maybe we should also explicitly stop calling the callback at line 16?

```js
@click="action.callback && !action.disabled ? action.callback() : () => {}"
```

At here, maybe we should create `handleCallback` function, and inside there, check `disabled` option and stop calling. Though it seems like it's already doing it... but...

ryo-gk:
It seems that the transition of the a tag is prioritized and so the callback is not executed...

lib/components/SCardHeader.vue
```diff
@@ -8,15 +8,23 @@
...
+              event
             >
               <component :is="getIcon(action.icon, !!action.disabled)" class="action-icon" />
-            </component>
+            </RouterLink>
+            <button
+              v-else
+              class="action"
+              :class="[action.mode || 'neutral', { disabled: action.disabled }]"
+              @click="action.callback()"
```
kiaking, line 24: We should stop calling the callback if `disabled: true`?

```suggestion
              @click="action.callback && !action.disabled ? action.callback() : () => {}"
```

PR #90, opened 2021-07-12
feat: add z-index for SToolTip

ryo-gk:
Currently, there is a possibility that the z-index o tooltips is overlapping with other components.
So make it possible to set a z-index for it.


kiaking (commented):
Maybe we should bump `--z-index-dropdown` to `1100` or `2000`? 🤔

kiaking (approved):

PR #93, opened 2021-07-14
feat: show footer only when there are actions more than one

ryo-gk:
Currently, the footer is displayed when no actions are passed,
but I think it is natural not to show the footer itself when no actions are passed.

### DIFF
--- a/lib/components/SCard.vue
+++ b/lib/components/SCard.vue
@@ -15,7 +15,7 @@
       <slot />
     </div>
 
-    <div v-if="footer" class="footer">
+    <div v-if="showFooter" class="footer">
       <SCardFooter :actions="footer.actions" />
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


lib/components/SCard.vue
```diff
@@ -44,6 +45,7 @@ export default defineComponent({
 
   setup(props) {
     const isCollapsed = ref(false)
+    const showFooter = computed(() => get(props.footer?.actions ?? []).length > 0)
```
kiaking, line 48: I think this should be

```suggestion
    const showFooter = computed(() => (get(props.footer?.actions) ?? []).length > 0)
```

So I guess below is a little more readable.

```suggestion
    const showFooter = computed(() => {
      const actions = get(props.footer?.actions) ?? []
      
      return actions.length > 0
    })
```

kiaking (changes requested):
@ryo-gk Well done! I made one comment 🙏

kiaking (approved):

PR #94, opened 2021-07-28
feat: implement button group component

ryo-gk:
sefirot_button-group


kiaking (approved):
Looking good! Adjusted a bit of styles, code format, English, etc.

PR #103, opened 2021-10-20
feat: add alignment option for text input and number input

ryo-gk:
Add `align` props to TextInput and NumberInput.
This is for the text alignment of the input and accepts 'left', 'right', and 'center' as a value.

image


ryo-gk:
Thank you for checking!
I fixed the reviewed points.

PR #108, opened 2021-11-15
feat(input-select): add error-message option for select input

ryo-gk:
This PR implements the `error-message` props for the SInputSelect component.


kiaking (approved):
Nice! Thank you!

PR #111, opened 2021-12-14
[next] fix(input number): emit null when empty is input

ryo-gk:
Fix the problem that emits 0 when empty is input.


kiaking (approved):
Niiiice!

PR #126, opened 2022-06-28
[next] feat(SInputText): add icon props to SInputText

ryo-gk:
Add icon props to SInputText component.


ryo-gk:
Implemented by the other PR.

PR #128, opened 2022-07-19
feat: Add histoire

ryo-gk:
Add histoire to preview the code of the components.


package.json
```diff
@@ -9,6 +9,9 @@
     "type": "vue-tsc --noEmit",
     "lint": "eslint --ext .js,.vue --ignore-path .gitignore --fix './{lib,tests}/**/*.{ts,vue}'",
     "lint:fail": "eslint --ext .js,.vue --ignore-path .gitignore './{lib,tests}/**/*.{ts,vue}'",
+    "story:dev": "histoire dev",
```
kiaking, line 12: Can we make this `story` (without `:dev`) instead? 👀

kiaking (approved):
Looking really gooooood! One comment. Other than that, I think this is ready to be merged.

package.json
ryo-gk: Yes, it looks better!

PR #138, opened 2022-09-20
feat: add callback option to STableCellText

ryo-gk:
Implement `onClick` option to STableCellText.

###  Usage
```ts
const table = {
  orders: ...,
  columns: [
    name: {
      label: 'Name',
      dropdown: dropdownName,
      cell: {
        type: 'text', 
        onClick: (value, record) => { ... }
      }
    }
  ]
}
```

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -79,6 +79,7 @@ export interface TableCellText extends TableCellBase {
   icon?: any
   value?: string | ((value: any) => string)
   link?(value: any, record: any): string
+  callback?(value: any, record: any): void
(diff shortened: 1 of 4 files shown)


lib/composables/Table.ts
```diff
@@ -79,6 +79,7 @@ export interface TableCellText extends TableCellBase {
   icon?: any
   value?: string | ((value: any) => string)
   link?(value: any, record: any): string
+  callback?(value: any, record: any): void
```
kiaking, line 82: Should we name this `onClick`? Since that's what we've been using in other places 👀

PR #141, opened 2022-09-21
feat: add icon component and support using iconify in button

### LINKED ISSUE #140: [Icon] Integrate antfu/icones or iconify
Currently, we place bear SVG component within `components/icons` folder. It would be cool if we can instead use [iconify](https://github.com/iconify/iconify) or [antfu/icones](https://github.com/antfu/icones).

I'm not sure what's the difference between antfu/icones and iconify. Probably, antfu/icones is using iconify, so iconify is what we want here...?

@brc-dd Could you investigate above packages and find a cool way to integrate it with sefirot? I don't know how those packages work, but what we want to do is import icons and use it in Vue Component.

It doesn't have to be integrated to Sefirot, and be peerDependencies of the project as well if that's how the package work. Or it can be exported from Sefirot like `sefirot/components/icons` or something.

### Notes

We need to be able to pass icons to Vue Component as Props.

```ts
import SomeIcon from 'cool-package'
    
<SButton :icon="SomeIcon />
```

brc-dd:
partially resolves #140

Migration path:

```diff
- import SIconImage from 'sefirot/components/icons/SIconImage.vue'
+ import iconImageSquare from '@iconify-icons/ph/image-square'

- <SButton :icon="SIconImage" />
+ <SButton :icon="iconImageSquare" />

# ---

- import SIconCheck from './icons/SIconCheck.vue'
+ import SIcon from './SIcon.vue'
+ import iconCheck from '@iconify-icons/ph/check'

- <SIconCheck class="check-icon" />
+ <SIcon :icon="iconCheck" class="check-iron" />

# after migration remove lib/components/icons
```

Custom icons can be defined like this:

```ts
export const customIcon = {
  // Icon body: <path d="..." />, required.
  body: '<path d="..." />',

  // optional:

  // Left position of viewBox.
  // Defaults to 0.
  left: 0, // number

  // Top position of viewBox.
  // Defaults to 0.
  top: 0, // number

  // Width of viewBox.
  // Defaults to 16.
  width: 24, // number

  // Height of viewBox.
  // Defaults to 16.
  height: 24 // number
}
```

Stuff in `SIcon.vue` can be added/customized later. Currently it just wraps the Iconify icon component.

🚨 `@iconify/vue@^4` is now a required peer dependency.

### DIFF
--- a/lib/components/SButton.vue
+++ b/lib/components/SButton.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { computed } from 'vue'
+import SIcon from './SIcon.vue'
 import SIconPreloader from './icons/SIconPreloader.vue'
 
 export type Size = 'mini' | 'small' | 'medium' | 'large' | 'jumbo'
@@ -21,7 +22,7 @@ const props = defineProps<{
   size?: Size
   type?: Type
   mode?: Mode
-  icon?: object
+  icon?: any
   label?: string
   href?: string
   loading?: boolean
@@ -61,7 +62,8 @@ function handleClick(): void {
     @click="handleClick"
   >
     <span class="content">
-      <span v-if="icon" class="icon"><component :is="icon" class="icon-svg" /></span>
+      <span v-if="icon && 'body' in icon" class="icon"><SIcon :icon="icon" class="icon-svg" /></span>
+      <span v-else-if="icon" class="icon"><component :is="icon" class="icon-svg" /></span>
       <span v-if="label" class="label">{{ label }}</span>
     </span>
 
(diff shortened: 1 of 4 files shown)


lib/components/SButton.vue
```diff
@@ -21,7 +22,7 @@ const props = defineProps<{
   size?: Size
   type?: Type
   mode?: Mode
-  icon?: object
+  icon?: any
```
kiaking, line 25: We can do this later, but maybe we could define this as `icon?: IconifyIcon`? And force users of the component to always pass in iconify icon (create a custom icon rather than passing in a plain SVG file if they need it).

I was wondering if we can avoid adding `v-if`s on every component receiving icon as a prop, but instead just simply do `SIcon :icon="icon` only.

Or, maybe we could make `SIcon` more abstruct and let it handle the `v-if` conditions.

kiaking: Wonderful! Thanks so much for the work 🙌

kiaking (approved):

PR #143, opened 2022-09-24
chore: setup spa redirects for histoire

brc-dd:
Currently opening a URL to a component directly (like https://story.sefirot.globalbrains.com/story/stories-components-savatar-story-vue?variantId=_default) or reloading the page after some client side redirection gives a 404. This should resolve it. It shouldn't conflict with VitePress' deployment because VP needs public directory inside `docs`.


kiaking (approved):
Ah! Thanks! Nice catch 👍

PR #144, opened 2022-09-24
refactor: replace all icons with iconify ones

### LINKED ISSUE #140: [Icon] Integrate antfu/icones or iconify

brc-dd:
closes #140

Note: We were (and are) using ["Feather Icons"](https://github.com/feathericons/feather). These aren't there on the Icones website. [The ones present there](https://icones.js.org/collection/fe) are ["Feather Icon"](https://github.com/feathericon/feathericon) (no trailing `s` 😅). So for searching use [the official tool](https://icon-sets.iconify.design/feather/) instead of Icones. (x-ref: https://github.com/antfu/icones/issues/93)

Also, IG the dependent projects will need to first migrate before merging this. Or we can probably restore the icons directory for now. And when all of them are using iconify icons, we can remove it.

Since feather icons are stroke-based (we had earlier modified them to be fill-based), there are certain changes to use `color` instead of `fill`. `fill: currentColor` is still present where icon is a prop, otherwise it's removed.


lib/components/SButton.vue
```diff
@@ -608,15 +608,16 @@ function handleClick(): void {
 
 .icon-svg {
   fill: currentColor;
-  transition: fill 0.25s;
+  transition-property: color, fill;
```
brc-dd, line 611: Earlier there were no transitions (color was being modified but fill was set to currentcolor, browsers don't animate in that case unless transition property is color). These changes will animate them, but I don't think we need these transitions. Open SButton >  Icon on Histoire (current deploy) and try changing the color mode (cmd+shift+d). Do the same for preview deploy.

kiaking:
Thanks a lot! I was thinking to go with [Phosphor](https://icones.js.org/collection/ph) and ditch Feather 👀 

> Also, IG the dependent projects will need to first migrate before merging this.

There's actually one, that currently uses this version of Sefirot, so I'll update that project as well. I'll merge this one first, and then work on with the project 👍 

I'll take over the PR from here 🙌 (including changes to Phosphor icons as well

kiaking:
I've migrated to `ph` icon instead and restored the PreLoader icon with the name `SSpinner`. For this spinner icon I wanted the previous look but couldn't find on icons.js, so... I just went ahead to create it as a component, rather than icon 😅

PR #146, opened 2022-09-26
fix(modal): listen on mousedown instead of click event

### LINKED ISSUE #145: [Modal] Handle "clickup" event correctly
When modal is closable by clicking the backdrop, currently it also handles clicked down on a different place, but clicked up on backdrop, and it's quite annoying so it would be cool if we can fix it.

For example, when we have following form opened as modal:

Screen Shot 2022-09-26 at 13 19 51

And if we want to select the whole text in the text input, we would click the end of the text in input, then drag it all the way left. At this point, if we drag it the whole way until the moush cursor is on the backdrop, then click up, modal closes 😓 

We would like to fix this to somehow detect both click down and up was triggered on backdrop.

Modal component can be found -> https://github.com/globalbrain/sefirot/blob/main/lib/components/SModal.vue

brc-dd:
fixes #145

For accessibility we can also add an event listener on keyup for escape.

### DIFF
--- a/lib/components/SModal.vue
+++ b/lib/components/SModal.vue
@@ -16,12 +16,12 @@ const emit = defineEmits<{
 
 const el = ref<any>(null)
 
-function onClick(e: any) {
+function onClick(e: MouseEvent) {
   if (!props.closable) {
     return
   }
 
-  if (e.target === el.value) {
+  if (e.button === 0 && e.target === el.value) {
     emit('close')
   }
 }
@@ -30,7 +30,7 @@ function onClick(e: any) {
 <template>
   <Teleport to="#sefirot-modals">
     <transition name="fade">
-      <div v-if="open" class="SModal" ref="el" @click="onClick">
+      <div v-if="open" class="SModal" ref="el" @mousedown="onClick">
         <slot />
       </div>
     </transition>


lib/components/SModal.vue
```diff
@@ -16,12 +16,12 @@ const emit = defineEmits<{
...
 const el = ref<any>(null)
 
-function onClick(e: any) {
+function onClick(e: MouseEvent) {
   if (!props.closable) {
     return
   }
 
-  if (e.target === el.value) {
+  if (e.button === 0 && e.target === el.value) {
```
kiaking, line 24: Just curious, what is `e.button`? 👀  Never used it before.

lib/components/SModal.vue
```diff
@@ -30,7 +30,7 @@ function onClick(e: any) {
 <template>
   <Teleport to="#sefirot-modals">
     <transition name="fade">
-      <div v-if="open" class="SModal" ref="el" @click="onClick">
+      <div v-if="open" class="SModal" ref="el" @mousedown="onClick">
```
kiaking, line 33: OK this was easy 😅 Nice. No idea why I couldn't come up with `mousedown` 😳

lib/components/SModal.vue
brc-dd: primary click (generally left button)

kiaking: Ah, gotcha! Nice to know 👍

kiaking:
Thanks a lot!

> For accessibility we can also add an event listener on keyup for escape.

Yeah we could, though I think we can do it in other PR. Maybe we need to be careful a bit. If we have for example dropdown menu input on modal, and maybe user might hit `esc` to intent to close the dropdown menu, instead of the whole modal.

Not sure how other modals out there handle such cases 🤔

brc-dd:
> maybe user might hit esc to intent to close the dropdown menu, instead of the whole modal

Yeah, I was just searching the same thing 😅. Directly adding event listener to document likely won't work.

kiaking:
Maybe each input must catch `esc` and prevent the bubble...? Not sure if `esc` bubbles but... Well anyway, let's just merge this one! 🚀

cuebit:
@kiaking all element events bubble. Though this is a user land decision, not something Sefirot will know about 👍

PR #147, opened 2022-09-26
chore: resolve tsconfig irregularities

cuebit:
This PR attempts to tackle some irregularities with `tsconfig.json` such as missing compiler files (notably vitepress config), including `stories` in the project and revising the consequential type issues.

Not sure if these were left out intentionally; nonetheless, the changes to `tsconfig.json` resolve issues that were not visible to the compiler that now is. See comments for more info.

### DIFF
--- a/lib/components/SButtonGroup.vue
+++ b/lib/components/SButtonGroup.vue
@@ -17,15 +17,14 @@
 <script setup lang="ts">
 import { computed, PropType } from 'vue'
 
-interface ButtonGroupItem {
+export interface ButtonGroupItem {
   label: string
   value: string
-  mode: Mode
+  mode?: Mode
 }
 
-type Mode = 'neutral' | 'info' | 'success' | 'warning' | 'danger'
-type Size = 'mini' | 'small' | 'medium' | 'large' | 'jumbo'
-
+export type Mode = 'neutral' | 'info' | 'success' | 'warning' | 'danger'
+export type Size = 'mini' | 'small' | 'medium' | 'large' | 'jumbo'
 
 const props = defineProps({
   items: { type: Array as PropType<ButtonGroupItem[]>, required: true },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,13 +6,12 @@
     "moduleResolution": "node",
     "strict": true,
     "isolatedModules": true,
-    "useDefineForClassFields": false,
     "jsx": "preserve",
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "sourceMap": true,
     "skipLibCheck": true,
-    "experimentalDecorators": true,
+    "noUnusedLocals": true,
 
     "lib": [
       "esnext",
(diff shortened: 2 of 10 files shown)


lib/components/SButtonGroup.vue
```diff
@@ -17,15 +17,14 @@
...
   value: string
-  mode: Mode
+  mode?: Mode
 }
 
-type Mode = 'neutral' | 'info' | 'success' | 'warning' | 'danger'
-type Size = 'mini' | 'small' | 'medium' | 'large' | 'jumbo'
-
+export type Mode = 'neutral' | 'info' | 'success' | 'warning' | 'danger'
+export type Size = 'mini' | 'small' | 'medium' | 'large' | 'jumbo'
```
cuebit, line 27: Export types for annotation in stories.

tsconfig.json
```diff
@@ -6,13 +6,12 @@
...
     "strict": true,
     "isolatedModules": true,
-    "useDefineForClassFields": false,
     "jsx": "preserve",
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "sourceMap": true,
     "skipLibCheck": true,
-    "experimentalDecorators": true,
+    "noUnusedLocals": true,
```
cuebit, line 14: Since the current eslint config skips detection of unused vars, this fallback prevents projects inheriting rogue types/assignments.

tsconfig.json
```diff
@@ -26,18 +25,22 @@
...
-    "env.d.ts",
-    "histoire.config.*",
-    "vitest.config.*"
+    "*.config.ts"
+  ],
+
+  "exclude": [
+    "node_modules",
+    "**/dist/**"
   ]
```
cuebit, line 45: Including `docs/.vitepress/**/*` ensures the compiler can pick up the files within the _hidden_ directory as `docs/.vitepress` alone does not get resolved. Up until now, `config.ts` was never visible to the compiler.

Excluding `dist` is no longer sufficient since we've offered the compiler a wildcard glob. `**/dist/**` ensures all nested dist directories (and there contents, in the case of `.vitepress`) are properly excluded.

Furthermore, it's safe to omit `env.d.ts` since it doesn't exist (or has been renamed for the time being). Can be added later at a more relevant time. `*.config.ts` removes the need to explicitly add individual files as they all share the same naming convention.

cuebit (commented):
Left some comments FYI.

tsconfig.json
kiaking: Ah, maybe we should bring back eslint rule...? I think it was turned off during the refactoring, I'm not sure. Can you check what errors do we get by turning back eslint rule on?

tsconfig.json
kiaking: Nice!

kiaking:
Great! Thanks Cue. It wasn't intentional, rather I was too lazy 😅

tsconfig.json
cuebit: Removed. It has no impact, but would have picked up some of the unused imports I've removed in this PR. 👍

tsconfig.json
cuebit: As a side, this cumbersome _include_ list results from `.vitepress`. A much cleaner resolution would be:

```json5
// tsconfig.json
{
  "include": ["."],
  "exclude": ["node_modules", "dist"]
}
```
And be explicit where it's needed:
```json5
// docs/tsconfig.json
{
  "extends": "../tsconfig.json",
  "include": ".vitepress/**/*"
}
```

This way or the revised version in this PR work just the same. 👍

tsconfig.json
cuebit: FYI @globalbrain/eslint-config ensures no mistakes are made. But that's another topic and PR. 👀

kiaking: Nice!

tsconfig.json
kiaking: Ah, clever 👀  But yeah I think it's OK to be explicit so. I would say the current version is fine 🚀

PR #149, opened 2022-09-29
Adjust some related to STable

ryo-gk:
Adjust / Fix the below points.
- Fix z-index of the STable header which was under the avatars.
- Fix empty pill was shown when value is null or undefined.
- Fix the border color of the STableCellAvatars on hover.
- Adjust to make it possible to accept Ref value for DropdownSectionFilterOption.

### DIFF
--- a/lib/components/SDropdownSectionFilter.vue
+++ b/lib/components/SDropdownSectionFilter.vue
@@ -11,7 +11,7 @@ import SIcon from './SIcon.vue'
 const props = defineProps<{
   search?: boolean
   selected: MaybeRef<DropdownSectionFilterSelectedValue>
-  options: DropdownSectionFilterOption[]
+  options: MaybeRef<DropdownSectionFilterOption[]>
   onClick?(value: string | number | boolean): void
 }>()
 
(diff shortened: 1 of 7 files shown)


lib/components/SDropdownSectionFilter.vue
```diff
@@ -95,6 +95,10 @@ function handleClick(option: DropdownSectionFilterOption, value: string | number
   border-bottom: 1px solid var(--c-divider-light);
   padding: 8px;
   background-color: var(--c-bg-elv-up);
+
+  .dark & {
```
kiaking, line 99: I don't think we need this. `--c-bg-elv-up` in dark mode is same as `--c-bg-mute` 👀

kiaking (commented):
Thanks! Added a comment 🙏

lib/components/SDropdownSectionFilter.vue
ryo-gk: Ah,  I'll fix it.

kiaking (approved):

PR #150, opened 2022-09-30
fix(table): prevent column dialog from going out of screen

### LINKED ISSUE #136: [Table] Column dialog gets cut off on mobile screen
On the mobile screen, the dialog gets cut off. We should make the dialog stick to either side of the screen in this case.

Screen Shot 2022-09-13 at 12 45 07

brc-dd:
fixes #136


ryo-gk:
Thanks!
It's looking good.

PR #163, opened 2022-10-20
fix(SInputDropdown): change border color on error

### LINKED ISSUE #160: [SInputDropdown] Error state missing border color
When an error happens on SInputDropdown, the border color doesn't turn red. I think we should make it red.

Screen Shot 2022-10-20 at 12 34 56

brc-dd:
fixes #160


kiaking (approved):
Ah damn this was easy. Nice 👍

PR #164, opened 2022-10-20
fix(SInputDropdown): prevent text wrapping in placeholder

### LINKED ISSUE #159: [SInputDropdown] Placeholder text gets wrapped
The placeholder on SInputDropdown gets wrapped when the text is longer than the input width. It should just get overflow hidden.

Screen Shot 2022-10-20 at 11 08 57

brc-dd:
fixes #159


kiaking (approved):
🚀 🚀 🚀

PR #165, opened 2022-10-20
fix: hide remove icon when only one item is selected and input is non nullable

### LINKED ISSUE #155: [SInputDropdown] Non nullable item still displays "remove icon"
Even the list item is not nullable, it shows "remove icon". We should hide close icon if the item is not removable (non-nullable).

👇 The "x" icon should not be displayed.

Screen Shot 2022-10-18 at 16 12 01

brc-dd:
fixes #155


kiaking (approved):
I went with passing down `removable` props to dropdown item components, since it required a bit more complicated handling depending on whether the modelValue is array or not 👍

PR #166, opened 2022-10-20
fix(SInputYMD): make `small` the default size

### LINKED ISSUE #156: [SInputYMD] Missing default value for `size` prop
Prop `size` should have `small` as a default but it's missing the default value.

brc-dd:
fixes #156


kiaking (approved):
Nice! Thanks!

PR #169, opened 2022-10-28
fix(dropdown): adjust spacing for text wrap

### LINKED ISSUE #168: [Dropdown] Text option text gets wrapped weirdly
I think we should wrap text instead of trimming it, but it should be wrapped beautifully.

Screen Shot 2022-10-28 at 15 52 29

brc-dd:
fixes #168 

image


kiaking (approved):
Thaaaaanks! Just adjusted a bit more on design side 👍

PR #170, opened 2022-10-31
feat: add custom component cell for STable

ryo-gk:
This PR implements the `component` option for STableCell's type, which can be used for adding custom cell to STable.

The usage is like below.

```vue
<script setup lang="ts">
const table = useTable({
  orders: [
    'message',
    ...
  ],

  columns: {
    message: {
      label: 'Message',
      cell: {
        type: 'component',
        component: markRaw(VTableCellMessage),
        props: {
          color: 'info',
          message: (value: any) => `${value}`
        }
      }
    }, ...
  },

  ...
})
</script>
```

```vue
// VTableCellMessage.vue (the component that you make)
<script setup lang="ts">
import { computed } from 'vue'

const props = defineProps<{
  color: 'info' | 'success'
  message: string | ((value: any) => string)
  value?: any
  record: any
}>()

const _message = computed(() => {
  return typeof props.message === 'function' ? props.message(props.value) : props.message
})
</script>

<template>
  <div class="VTableCellMessage">
    Someone says "{{ _message }}"
  </div>
</template>
```

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -82,7 +82,7 @@ export interface TableCellAvatars extends TableCellBase {
 export interface TableCellComponent extends TableCellBase {
   type: 'component'
   component: Component
-  props: Record<string, any>
+  props?: Record<string, any>
 }
 
 export interface TableCellAvatarsOption {
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -1,5 +1,5 @@
 import { MaybeRef } from '@vueuse/core'
-import { reactive } from 'vue'
+import { Component, reactive } from 'vue'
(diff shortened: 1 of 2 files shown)


lib/composables/Table.ts
```diff
@@ -78,6 +79,12 @@ export interface TableCellAvatars extends TableCellBase {
   color?: 'neutral' | 'soft' | 'mute'
 }
 
+export interface TableCellComponent extends TableCellBase {
+  type: 'component'
+  component: Component
+  props: Record<string, any>
```
kiaking, line 85: I think this should be optional since some component might not require any props 👀

kiaking: Nice 👍

PR #176, opened 2022-12-15
feat(table): allow callback on cell property

### LINKED ISSUE #174: [Table] Accept call back on cell property
We have many use cases where we need to do some checks on cell rendering using the current data it's passed. Now we can already do that for few cell fields such as `value` or `color`, but sometimes we need to do the same mutation in several places. It would be great if we can simply make `cells` accept callback.

```ts
// Current type
export interface TableColumn {
  label: string
  className?: string
  dropdown?: DropdownSection[]
  cell?: TableCell  // <- Change this
}

// This would be great
export interface TableColumn {
  label: string
  className?: string
  dropdown?: DropdownSection[]
  cell?: TableCell | ((value: any, record: any) => TableCell)
}
```

Actual usage is:

```ts
const table = useTable({
  ...
  
  columns: {
    label: 'name',
    cell(value, record) {
      // Reuse this calculation on both `value` and `color`.
      const status = getStatus(record)

      return {
        type: 'text',
        value: status,
        color: status === 'draft' ? 'mute' : 'neutral'
      }
    }
  }
})
```

brc-dd:
fixes #174

### DIFF
--- a/lib/components/STableCell.vue
+++ b/lib/components/STableCell.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { computed } from 'vue'
 import { TableCell } from '../composables/Table'
 import STableCellAvatar from './STableCellAvatar.vue'
 import STableCellAvatars from './STableCellAvatars.vue'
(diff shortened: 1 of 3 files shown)


NozomuIkuta (commented):
@brc-dd 

Thank you for the PR!

Would you modify `stories/components/STable.story.vue` so that we can see your PR working in Histoire?

NozomuIkuta (approved):
@brc-dd 

Thanks!
LGTM and I will leave it to @kiaking!

kiaking (approved):
Thanks a bunch! I've reverted the story change since it is bit hard to know the difference unless you know the internal code 😅 

But other than that, awesome as always!

PR #179, opened 2022-12-22
feat: add STableCellPills component

### LINKED ISSUE #177: [Table] Error prop type on Cell component because of external type
We're getting this error.

```
[Vue warn]: Invalid prop: type check failed for prop "cell". Expected Null | Function, got Object  
  at <STableCell name="status" class-name=undefined cell= Object  ... > 
  at <STableItem key="status" name="status" class-name=undefined  ... > 
```

I think this is because we made changes in #174 and now `cell` props is defined as `TableCell | ((value: any, record: any) => TableCell)` which becomes `null | Function` because of current script setup limitation 😢 

I guess we need to redeclare TableCell type inside `STableCell` component.

### LINKED ISSUE #178: [Table] Add STableCellPills component
We need a STableCellPills component that accepts `string[]` as `value`, and show them as pills.
This is because the `STableCellPill` which we already have can't show multiple pills.

Design image is below. ("Topic type" column)
image

The usage is like below.

```ts
const table = useTable({
  ...
  columns: {
    tags: {
      label: 'Tags',
      cell: {
        type: 'pills',
        pills(items, record) {
          return items.map((item) => {
            return {
              label: item,
              color: item === 'info' ? 'info' : 'mute'
            }
          })
        }
      },
  },
  records: [
    { ... , tags: ['Info', 'News'], ... },
    { ... , tags: ['Info'], ... },
  ],
  ...
})
  ...
})
```

~Or, might be simpler to allow `STableCellPill` to accept also `string[]`?~

brc-dd:
fixes #178
fixes #177

### DIFF
--- a/lib/components/STableCell.vue
+++ b/lib/components/STableCell.vue
@@ -5,18 +5,19 @@ import STableCellAvatar from './STableCellAvatar.vue'
 import STableCellAvatars from './STableCellAvatars.vue'
 import STableCellDay from './STableCellDay.vue'
 import STableCellPill from './STableCellPill.vue'
+import STableCellPills from './STableCellPills.vue'
 import STableCellText from './STableCellText.vue'
 
 const props = defineProps<{
   name: string
   className?: string
-  cell?: TableCell | ((value: any, record: any) => TableCell)
+  cell?: any
   value: any
   record: any
   records: Record<string, any>
 }>()
 
-const cell = computed(() =>
+const computedCell = computed<TableCell | undefined>(() =>
   typeof props.cell === 'function'
     ? props.cell(props.value, props.record)
     : props.cell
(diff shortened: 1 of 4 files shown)


stories/components/STable.story.vue
```diff
@@ -85,35 +100,67 @@ const dropdownCreatedAt = createDropdown([
...
+    createdAt: '2022-09-08',
+    tags: ['Info']
+  }
 ]
 
 const filteredData = computed(() => {
   return data
     .filter((i) => filterBy(i.status, dropdownStatusSelected.value))
     .filter((i) => filterBy(i.type, dropdownTypeSelected.value))
+    .filter((i) => i.tags.some((tag) => filterBy(tag, dropdownTagsSelected.value)))
```
brc-dd, line 155: Might wanna modify this. I'm not completely sure what's the expected behavior is with filtering multiple items.

lib/components/STableCell.vue
```diff
@@ -5,18 +5,19 @@ import STableCellAvatar from './STableCellAvatar.vue'
...
   className?: string
-  cell?: TableCell | ((value: any, record: any) => TableCell)
+  cell?: any
   value: any
   record: any
   records: Record<string, any>
 }>()
 
-const cell = computed(() =>
+const computedCell = computed<TableCell | undefined>(() =>
```
brc-dd, line 20: Had to rename this because Volar was getting confused in the template.

brc-dd:
Also there are a bunch of warning/errors on the console when someone filters something 👀  (they were before this PR too).

ryo-gk:
Thanks a lot 🙌 
Looking good to me!

@kiaking 
Could you check, and tweak some designs if you need?

lib/components/STableCell.vue
kiaking: Ah, gotcha. No problem 👍

stories/components/STable.story.vue
kiaking: Ah, well this stuff depends on the business logic anyway so... Let's come back if we find issues on it 👀

kiaking:
> Also there are a bunch of warning/errors on the console when someone filters something

That's the one coming from Histoire I think... not sure what it is. It doesn't happen on actual app that uses STable component. It would be cool if we can fix it though 🤔

PR #182, opened 2023-01-08
fix(input-select): stylings not applied correctly

kiaking:


brc-dd:
On Windows it's looking like this:

image

Looks fine on mac though. On Windows IG it's applying the text color to options too.

kiaking:
@brc-dd Thanks for checking! I think you're right, seems like `color` specified at `<select>` is inherited in `<option>` too, but not sure how to fix it 🤔 

So, it will look OK on `light` mode... right? Because then the text color should be black.

I've set `<option>` color to `color: initial`... It might help? Could you test it and see if it made any difference 👀 ?

brc-dd:
> I've set <option> color to color: initial... It might help?

Yeah, it works!

kiaking:
Yay! 🎉  Thanks for checking 👍 

Maybe for VitePress we might want to wrap `<Showcase>` with `vp-raw` 🤔

brc-dd:
Yeah that or whyframe would work. 👍

PR #188, opened 2023-01-14
fix(select): use index as option value

### LINKED ISSUE #183: Make `SInputSelect` component test coverage 100%
I couldn't find out how to trigger `select` value change with Vue Test Utils. Tried `setValue` but couldn't make it work properly.

It would be cool if some one can tackle this 🤔  We need to test the `change` event on `select` component and see if component emits the correct value.

brc-dd:
fixes #183


kiaking:
Smart! Nice idea 👍

PR #193, opened 2023-01-17
test(select): add tests for `value`, `modelValue`, `placeholder` props

### LINKED ISSUE #191: Make `SInputSelect` test coverage 100%
It's not yet there.

- [Current Test Coverage](https://app.codecov.io/gh/globalbrain/sefirot/blob/main/lib/components/SInputSelect.vue)

For testing `value` and `modelValue` prop, we have similar test in other input components like `SInputNumber`.

```ts
test('it accepts `value` prop', async () => {
  const wrapper = mount(SInputNumber, {
    props: { value: 1 }
  })

  expect(getInputValue(wrapper.find('.SInputNumber .input'))).toBe('1')
})

test('it sets value to `null` when nither `value` or `modelValue` is set', async () => {
  const wrapper = mount(SInputNumber)

  expect(getInputValue(wrapper.find('.SInputNumber .input'))).toBe('')
})
```

brc-dd:
fixes #191

### DIFF
--- a/lib/components/SInputSelect.vue
+++ b/lib/components/SInputSelect.vue
@@ -56,14 +56,6 @@ function isSelectedOption(option: Option): boolean {
   return option.value === _value.value
 }
 
-function focus() {
-  isFocused.value = true
-}
-
-function blur() {
-  isFocused.value = false
-}
-
 function emitChange(e: any): void {
   if (!props.disabled) {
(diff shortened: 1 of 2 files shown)


kiaking (approved):
Thanks! Will merge this as a fix of component behaviors since it does actually fix the undefined `value` prop problem 😳

PR #195, opened 2023-01-17
test: add tests for SInputBase

### LINKED ISSUE #192: Make `SInputBase` test coverage 100%
- [Current Test Coverage](https://app.codecov.io/gh/globalbrain/sefirot/blob/main/lib/components/SInputBase.vue)

brc-dd:
fixes #192 

(it's _slightly_ better, but isn't 100% though)

### DIFF
--- a/docs/components/input-file.md
+++ b/docs/components/input-file.md
@@ -159,20 +159,24 @@ interface Props {
 
 ### `validation`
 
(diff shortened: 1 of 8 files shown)


kiaking:
Yeah indeed this was impossible to test, since `errorMsg` computed would never be called without `validation` being present 👀 

I've refactored `SInputBase` to use single computed and returning object. Also, refactored `Validatable` type and added `createValidatable` test utils for future testing 👍

PR #198, opened 2023-01-19
feat: add `maxFileSize` validation rule

### LINKED ISSUE #197: Add `maxFileSize` validation rule
## Validator

Implement the `fileSize` validator. It would be cool if we can specify values like `KB`, `MB` and such.

```ts
function maxFileSize(file: File, size: string): boolean
```

```ts
maxFileSize(file, '500KB')
maxFileSize(file, '100MB')
```

## Rule

Add `maxFileSize` rule.

```ts
/**
 * Default error message:
 * `The file must be less or equal to ${$params.size}.`
 */
function maxFileSize(size: string, msg?: string): ValidationRuleWithParams
```

brc-dd:
fixes #197


kiaking (commented):
Looking good! Can we have test for this? 👀

brc-dd:
This won't work with decimal in size, do you want to process that too? Also, it won't work with IEC units (KiB, MiB ... power of 2**10).

kiaking:
Nah that's fine! I think for now we only need `KB`, `MB` and such 👍

kiaking (approved):
Thanks a lot!

PR #203, opened 2023-01-24
feat(table): add `resizable` prop

### LINKED ISSUE #201: Add option to disable column resize on specific column in `STable`
It would be cool to have a option to disable column resize on table, not all but for specified column.

The use case is when we have big table with lots of column, adding empty last column makes the UI a bit nicer similar to what GitHub project has.

In this case, we wanna disable resizing the last empty column because there's no point of doing it, and it always allow table to have good background color (if we resize the last item, we get black bg color).

API idea.

```ts
const table = useTable({
  columns: {
    lastItem: {
      resizable: false
    }
  }
})
```

brc-dd:
fixes #201

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -182,17 +182,32 @@ const options = useTable({
 })
 ```
 
+## Disable column resize
+
+By default. all columns are resizable via the user. You may disable resizing for each column you specify by setting `resizable` option.
+
+```ts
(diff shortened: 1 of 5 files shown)


kiaking:
Niiice! Looking good 👍  I'll add docs and merge this one 🚀

brc-dd:
I actually forgot that we have docs now 😅

kiaking:
Haha, yeah I've added the table docs just recently so 👀

PR #204, opened 2023-01-24
feat: show dropdown on top if input is near viewport bottom

### LINKED ISSUE #200: Auto position dropdown dialog for `SInputDropdown`
Currently, the dropdown dialog of `SInputDropdown` is always pops up at the bottom. It would be grate if it shows up above when there's no space under the input. This happens a lot when you have `SInputDropdown` at the last item in the form.

Additionally, maybe it would make sense to add `position: 'top' | 'bottom'` prop to explicitly set the position.

brc-dd:
fixes #200

Demo (doesn't seem smooth though 👀):

https://user-images.githubusercontent.com/40380293/214350417-a6d5e26c-a627-48ae-8f69-106c3b383694.mov

### DIFF
--- a/lib/components/SInputDropdown.vue
+++ b/lib/components/SInputDropdown.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import IconCaretDown from '@iconify-icons/ph/caret-down-bold'
 import IconCaretUp from '@iconify-icons/ph/caret-up-bold'
+import { useElementBounding, useWindowSize } from '@vueuse/core'
 import xor from 'lodash-es/xor'
 import { computed } from 'vue'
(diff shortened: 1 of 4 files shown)


kiaking:
Looking great! I've updated the code to only calculate the position on the open event and stay the same during the scroll. I think it's a bit more intuitive for the user... maybe. But it's good to know making it work on scroll is easy as pie with vueuse!

Also, I've added a limit to the upper space. So, if we don't have enough space at the top, display the dialog at bottom always. This might happen a lot in mobile device or when user have quite short window size.

Could you review my changes and see if didn't do anything stupid here 👀

brc-dd:
Yeah seems fine 👍

kiaking (approved):

kiaking:
Thanks!

PR #205, opened 2023-01-25
chore: enable corepack

NozomuIkuta:
I found that no package manager and its version are specified in the code base.

I believe it's reasonable to specify `packageManager` field in `package.json` and to use corepack, so that everyone can use same version of same package manager always.

I set everyone as reviewers to hear what you think, or just notification purpose.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,8 +14,6 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v2
-        with:
-          version: 7.0.0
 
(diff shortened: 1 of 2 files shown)


brc-dd:
Might wanna remove pnpm version from test workflow too.

kiaking (approved):

cuebit (approved):

brc-dd (approved):

ryo-gk (approved):

kiaking:
I've removed pnpm version from GitHub workflow 👍

PR #206, opened 2023-01-27
feat(input): add `addon-*` slots

### LINKED ISSUE #202: Add "Addon Before" feature to `SInputText` and `SInputNumber`
design [Figma Design](https://www.figma.com/file/V9aRWurAXc6JggujTa2J4p/GB-Design-Kits?node-id=47%3A38&t=ou2UFLMkjDquZRi7-1)

The "Addon" is the feature where users can add an extra label or action button to the input.

Screen Shot 2023-01-24 at 13 30 17

## Basics

Add `:addon-before` prop to `SInputText` and `SInputNumber` components. We add `-before` suffix because probably we want to have `:addon-after` in the future.

`:addon-before` can take a few types of value. `string`, `number`, or an icon. It should also have `addon-before-clickable` prop to make the addon act as a button.

```ts
import { IconifyIcon } from '@iconify/vue/dist/offline'

interface Props {
  // The label for the addon.
  addonBefore?: string | number | IconifyIcon

  // Whether the addon act as button. Defaults to `false`.
  addonBeforeClickable?: boolean
}
```

When `addonBeforeClickable` is true, it emits an event when the user clicks on it.

```ts
interface Emits {
  // Emits only when `Props.addonBeforeClickable === true`.
  (e: 'addon-before:click'): void
}
```

When adding this prop, it should display a component that looks 👇 

Screen Shot 2023-01-24 at 13 30 17

## Dropdown feature

Provide a way to add a dropdown feature to the addon as well. When this feature is enabled, users may click the addon, and display the `SDropdown` component as `SInputDropdown` does.

```ts
import { DropdownSectionFilter } from '@/composables/Dropdown'

interface Props {
  addonBefore?: string | number | IconifyIcon
  
  // Dropdown option same as `SInputDropdown`.
  addonBeforeDropdown?: AddonDropdownOption[]

  // Disable the dropdown search option. Defaults to `false`.
  addonBeforeDropdownNoSearch?: boolean

  // Whether to close the dialog when a user selects an item.
  // Defaults to `false`.
  addonBeforeDropdownCloseOnClick?: boolean
}

type AddonDropdownOption =
  | AddonDropdownOptionText
  | AddonDropdownOptionAvatar

interface AddonDropdownOptionBase {
  type?: 'text' | 'avatar'
  value: OptionValue
}

type AddonDropdownOptionValue = string | number | boolean

interface AddonDropdownOptionText extends AddonDropdownOptionBase {
  type?: 'text'
  label: string
}

interface AddonDropdownOptionAvatar extends AddonDropdownOptionBase {
  type: 'avatar'
  label: string
  image?: string | null
}
```

When dropdown is enabled, emit `addon-before:click` even when `addon-before-clickable` is not `true`. Also, emits additional events for dropdown item selection.

```ts
interface Emits {
  // Always emits when dropdown is enabled.
  (e: 'addon-before:click'): void

  // Emits when a user selects the item in dropdown.
  (e: 'addon-before:change', value: AddonDropdownOptionValue): void
}
```

When dropdown is enabled, it should show down caret after the addon label.

Screen Shot 2023-01-24 at 13 37 36

## Disbale

The addon should check for `disabled` prop and if it's `true`, stop emitting any events. The design would should look like this.

Screen Shot 2023-01-24 at 13 37 41

## Light mode appearance

Screen Shot 2023-01-24 at 13 37 59

NozomuIkuta:
Asana: https://app.asana.com/0/1130160932501704/1203800531766512/f
Resolves #202

I believe I covered all the necessary use cases in Histoire.

### DIFF
--- a/docs/components/input-addon.md
+++ b/docs/components/input-addon.md
@@ -1,0 +1,178 @@
+<script setup lang="ts">
+import { ref } from 'vue'
+import SInputAddon from 'sefirot/components/SInputAddon.vue'
+import SInputText from 'sefirot/components/SInputText.vue'
+
+const input = ref<string | null>(null)
+</script>
+
+# SInputAddon
+
+`<SInputAddon>` is a special component that can be used to add extra label or action to other input.
+
+<Showcase
+  path="/components/SInputText.vue"
+  story="/stories-components-sinputtext-02-addons-story-vue"
+>
+  <SInputText placeholder="johndoe" v-model="input">
+    <template #addon-before>
+      <SInputAddon label="@" :clickable="false" />
+    </template>
+  </SInputText>
+</Showcase>
+
+## Usage
+
+In order to use `<SInputAddon>`, you must inject it through slots. Currently, the supported components are listed below.
+
+- `<SInputText>`
+- [`<SInputNumber>`](/components/input-number)
+
+You may choose the position of the addon to be injected using the slot name either `#addon-before` or `#addon-after`.
+
+```vue
+<script setup lang="ts">
+import { ref } from 'vue'
+import SInputAddon from '@globalbrain/sefirot/lib/components/SInputAddon.vue'
+import SInputText from '@globalbrain/sefirot/lib/components/SInputText.vue'
+
+const input = ref<string | null>(null)
+</script>
+
+<template>
+  <SInputNumber placeholder="johndoe" v-model="input">
+    <!-- Inject addon before the input box. -->
+    <template #addon-before>
+      <SInputAddon label="@" />
+    </template>
+
+    <!-- Inject addon after the input box. -->
+    <template #addon-after>
+      <SInputAddon label="@" />
+    </template>
+  </SInputNumber>
+</template>
+```
+
+## Props
+
+Here are the list of props you may pass to the component.
+
+### `:label`
+
+Defines the label of the addon button.
+
+```ts
+import { IconifyIcon } from '@iconify/vue/dist/offline'
+
+interface Props {
+  label?: string | IconifyIcon
+}
+```
+
+```vue-html
+<SInputAddon label="@" />
+```
+
+### `:clickable`
+
+Defines whether the button should be clickable. Defaults to `true`.
+
+```ts
+interface Props {
+  clickable?: boolean
+}
+```
+
+```vue-html
+<SInputAddon label="@" :clickable="false" />
+```
+
+### `:dropdown`
+
+Defines dropdown option. When this prop is set, the dropdown will pop up when user clicks the addon button. In addition, if `:label` is not defined, the selected option's label will be used for label value.
+
+```ts
+import { DropdownSection } from '@globalbrain/sefirot/lib/composables/Dropdown'
+
+interface Props {
+  dropdown?: DropdownSection[]
+}
+```
+
+```vue
+<script setup lang="ts">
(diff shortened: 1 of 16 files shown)


lib/components/SInputAddon.vue
```diff
@@ -0,0 +1,133 @@
...
+
+function getPosition() {
+  if (props.dropdowpPosition) {
+    return props.dropdowpPosition
+  }
+
+  const dialogHeight = 400
+
+  // If the space top of the input is not enough to show dialog, just show
+  // the dialo at the bottom of the input.
```
ryo-gk, line 90:
```suggestion
  // the dialog at the bottom of the input.
```

Maybe typo

stories/components/SInputNumber.02_Addons.story.vue
```diff
@@ -0,0 +1,286 @@
...
+            separator
+          >
+            <template #addon-before>
+              <SInputAddon label="¥" unclickable />
+            </template>
+          </SInputNumber>
+
+          <div class="actions">
+            <SButton size="small" mode="info" label="Submit" @click="validateAndNotify" />
+            <SButton size="small" mode="mute" label="Reset" @click="init" />
```
ryo-gk, line 114: It may be better that it is enough to have just one button section of `Submit` and `Reset`, although now there are for each `TextInput` component, since the function for them is common.

stories/components/SInputNumber.02_Addons.story.vue
```diff
@@ -0,0 +1,286 @@
...
+            <template #addon-after>
+              <SInputAddon :label="IconDotsThree" :dropdown="dropdown2" no-dropdown-arrow />
+            </template>
+          </SInputNumber>
+        </div>
+      </div>
+      <div class="group">
+        <div class="title">Addon Dropdown (Before / After / Disabled)</div>
+        <div class="grid">
+          <SInputText
```
ryo-gk, line 239: This is a typo for `SInputNumber`, and so are the props.

ryo-gk (commented):
@NozomuIkuta Added some comment. 
Please check them🙏

stories/components/SInputNumber.02_Addons.story.vue
kiaking: I've removed these ones 👍

stories/components/SInputNumber.02_Addons.story.vue
kiaking: Fixed!

kiaking (approved):
Thanks a lot! Looking good. Adjusted few stuff here and there.

- Added tests and docs.
- Changed props default to be true (e.g. unclickable -> clickable).
- Cleaned up stories.
- Cleaned up styles along with `<SInputText>`.
- Moved some functions into `Dropdown` composable.

docs/components/input-addon.md
```diff
@@ -0,0 +1,178 @@
...
+}
+```
+
+```vue-html
+<SInputAddon :dropdown="dropdown" :dropdown-position="bottom" />
+```
+
+### `:disabled`
+
+Disable the addon action. When this prop is set, no events ets emitted.
```
NozomuIkuta, line 154:
```suggestion
Disable the addon action. When this prop is set, no events are emitted.
```

kiaking: Thanks! Fixed 👍

PR #210, opened 2023-01-31
feat: add `unit-*` props to `SInputText` and `SInputNumber`

NozomuIkuta:
I added `unit-before` and `unit-after` props to `SInputText` and `SInputNumber`.

Additionally, I modified Histoire's story controls for us to test UI interactively.


kiaking (approved):
Woot woot 🙌

PR #215, opened 2023-01-31
feat(input-date): add `block` option

### LINKED ISSUE #212: Add `block` option to `SInputDate`
Currently, the width of `SInputDate` is fixed to the smallest possible size. But in some case, it would be nice for the input to expand fully relative to parent container. So, let's add this option!

## Props

```ts
interface Props {
  block?: boolean // Default `false`.
}
```

```html
<SInputDate block />
```

brc-dd:
fixes #212

### DIFF
--- a/lib/components/SInputDate.vue
+++ b/lib/components/SInputDate.vue
@@ -9,6 +9,7 @@ import SInputBase from './SInputBase.vue'
 export type Color = 'neutral' | 'mute' | 'info' | 'success' | 'warning' | 'danger'
 
 const props = defineProps<{
+  block?: boolean
   name?: string
   label?: string
   info?: string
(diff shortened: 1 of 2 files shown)


kiaking (approved):
Thanks a bunch! 🎉 🎉 🎉

lib/components/SInputDate.vue
```diff
@@ -71,6 +72,7 @@ function emitBlur() {
         <input
           :id="name"
           class="input"
+          :class="{ block }"
```
brc-dd, line 75: Ah, this is nice 😅

kiaking: Oh yeah 😳

PR #218, opened 2023-02-01
feat: add  `text-color` props to SInputText and SInputNumber

ryo-gk:
Add the `text-color` props which change the color of the input value of `SInputText` and `SInputNumber`.


kiaking (approved):
Wonderful!

PR #222, opened 2023-02-14
feat(button): enhance `SButtonGroup` interface

NozomuIkuta:
Asana: N/A

This PR is to enhance `SButtonGroup` component in the following points:

- Add `Icon` property to `ButtonGroupItem` interface, which is prioritized than `label` if both exist
- Add `disabled` property to `ButtonGroupItem` interface
- Add `beforeText` and `afterText` props to show some computed text (e.g. current page)
- Add `click` event, which is emitted with `value` of the clicked button, so that `beforeText` and/or `afterText` can be calculated on the fly
- Modernize some lines of code (e.g. `defineProps` with TypeScript)
- Fix some styles (e.g. overflown inner height)


kiaking:
We've discussed this internally and made changes accordingly.

PR #229, opened 2023-03-16
feat: add size props to SInputDate

ryo-gk:
This PR adds `size` prop to SInputDate.
Basically the style for each sizes is same as `SInputText`.


kiaking (approved):
Niiice!

PR #231, opened 2023-03-27
feat(table): allow adding summary row

### LINKED ISSUE #230: [Table] Add `summary` feature
It would be cool add `summary` row in the table. Looks like this (The "Total" row).

Screenshot 2023-03-27 at 13 35 54

## API

This field doesn't have to be auto calculated. Since we could get this data from the server side for example. So, at least for the start, we can create this records manually. So, maybe we could add `summary` field?

```ts
const table = useTable({
  orders: [...],
  columns: { ... },

  summary: {
    name: 'Total',
    type: null,
    estTotalAmount: '267,870.10',
    acqShares: 'N/A',
    acpTotalAmount: '20,500.00',
  }
})
```

The `summary` should use the settings defined on `columns` as same as other field. Except that by default, it will set `font-weight: 600` as a style. Other than that it should work as same as normal records.

## Keep in mind

The row should be appended right after the normal records so that it gets horizontal scrolling.

brc-dd:
fixes #230

- Sorting and filtering does not affect the summary row (always shown).
- ~~Pagination is not working (at least not on Histoire), so not sure how it will play with that 👀~~
  - `onPrev`/`onNext` is empty there, will check if it works fine or not

Also, on filtering, there are a bunch of errors (not related to this PR):

image

TODO:
- [x] Update docs
- [x] Add tests

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -34,10 +34,6 @@ export type TableColumns<
     : TableColumn<undefined, R>
 }
 
-export type TableSummary<T> = {
-  [P in keyof T]?: T[P] | null | undefined
-}
-
 export interface TableColumn<V, R> {
   label?: string
   className?: string
@@ -141,6 +137,10 @@ export interface TableCellComponent extends TableCellBase {
   type: 'component'
   component: Component
   props?: Record<string, any>
+}
+
+export type TableSummary<T> = {
+  [P in keyof T]?: T[P] | null | undefined
 }
 
 export interface UseTableOptions<O extends string, R extends Record<string, any>> {
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -4,6 +4,10 @@ import { reactive } from 'vue'
 import type { Day } from '../support/Day'
 import type { DropdownSection } from './Dropdown'
 
+type Optional<T> = {
+  [P in keyof T]?: T[P] | null | undefined
+}
+
 export interface Table<
   O extends string = any,
   R extends Record<string, any> = any
@@ -13,6 +17,7 @@ export interface Table<
   records?: R[] | null
   header?: boolean
   footer?: boolean
+  summary?: Optional<R> | null
   total?: number | null
   page?: number | null
   perPage?: number | null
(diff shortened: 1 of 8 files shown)


brc-dd:
I though of adding it to records itself, but that will make it sortable. Although I think we can still avoid duplicate code for row by using a computed property instead. Also, for some reason pagination isn't working (at least on Histoire), so not sure how it will play with that 👀

lib/composables/Table.ts
```diff
@@ -4,6 +4,10 @@ import { reactive } from 'vue'
 import type { Day } from '../support/Day'
 import type { DropdownSection } from './Dropdown'
 
+type Optional<T> = {
```
kiaking, line 7: Let's make this public and name `TableSummary` to be consistent with other stuff like `TableColumns`.

Just so that it looks good when listing this type in the docs 👍 (Although we don't have a docs for this composable yet but for the future 👀

kiaking (approved):
Looking nice! Very smart 👍

kiaking:
> Pagination is not working (at least not on Histoire), so not sure how it will play with that 👀
> - onPrev/onNext is empty there, will check if it works fine or not

Yap! Currently updating records are manual so, if the summary should change on page change, it should be done at user land. Just pass reactive value to `summary` option and update it. Though, I think it's rare to use summary for table with pagination, but I don't see any issues.

> Also, on filtering, there are a bunch of errors (not related to this PR):

hmmmm something happening on Histoire I guess 🤔 ...

lib/composables/Table.ts
brc-dd: What should be the `TableSummary`?

```ts
export type TableSummary<R> = Optional<R>
```

This? Or did you mean change Optional to TableSummary:

```ts
export type TableSummary<T> = {
  [P in keyof T]?: T[P] | null | undefined
}
```

kiaking: Either way is fine! But for now, The later is what I was thinking 👍 So this.

```ts
export type TableSummary<T> = {
  [P in keyof T]?: T[P] | null | undefined
}
```

brc-dd: Done! 🙌

kiaking: 🎉

kiaking (approved):
Thanks a bunch! I've adjusted the style and defined new css var for font weight.

PR #240, opened 2023-04-06
feat(table): add an option to grow a column to fill available space

### LINKED ISSUE #239: [Table] Make `--table-col-max-width` really work
When creating a table, there're many cases where we wanna set the first column to "Set the min width, but fill up empty space if there's any". For this reason, we have `--table-col-max-width`. However, this only works if the content size is smaller than the `min-width`. Here is the broken example.

Screenshot 2023-04-06 at 11 13 01

For this example, the first column has following css defined.

```css
.table :deep(.col-name) {
  --table-col-width: 160px;
  --table-col-max-width: 100%;
}
```

With this definition, I wand the column to have min of 160px, but if the window is large, then fill up the rest of the space. Similar to how `flex-grow: 1` would work.

Currently `--table-col-max-width` sets the max width to 100%. So, if the content is larger than `160px` (min width), it just gets bigger.

The hard part is that the table can be overflow. If there're lots of column, then it should exceed the table width and should be horizontally scrollable.

Is there a way to achieve this?

brc-dd:
fixes #239

https://user-images.githubusercontent.com/40380293/230400168-5dbfe1d4-7603-425e-a343-6b01f009dcf2.mov

#### TODO:
- [x] handle column resizing smoothly
- [x] cleanup code and think of better prop names and pr title 😅
- [x] update docs

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -39,7 +39,6 @@ export interface TableColumn<V, R, SV, SR> {
   className?: string
   dropdown?: DropdownSection[]
   grow?: boolean
-  fillOnAdjust?: boolean
   resizable?: boolean
   cell?: TableCell | TableColumnCellFn<V, R>
   summaryCell?: TableCell | TableColumnCellFn<SV, SR>
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -38,6 +38,7 @@ export interface TableColumn<V, R, SV, SR> {
   label?: string
   className?: string
   dropdown?: DropdownSection[]
+  grow?: boolean
   resizable?: boolean
   cell?: TableCell | TableColumnCellFn<V, R>
   summaryCell?: TableCell | TableColumnCellFn<SV, SR>
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -39,6 +39,7 @@ export interface TableColumn<V, R, SV, SR> {
   className?: string
   dropdown?: DropdownSection[]
   grow?: boolean
+  fillOnAdjust?: boolean
   resizable?: boolean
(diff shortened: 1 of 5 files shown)


brc-dd:
Not so sure about the last commit. Do we want to stop autosizing when any of the columns is adjusted?

kiaking:
> Not so sure about the last commit. Do we want to stop autosizing when the width of any of the columns is adjusted? (It'll behave like this then:)

Good point! Hmmm... I guess so. The video looks good and I like it.

Also, if we were to not do that, which is we try to only stop autosizing when the auto sizable column is adjusted, it's bit hard to tell when would that be. Is that when user drags the right border of the column or left...?

This is hard to tell from user point of view, so stop auto sizing all together whenever any column gets resized, make sense, I think 🤔 

But, I don't think this has a huge impact on usability, so I say let's choose the "easier to implement" option 😄

lib/composables/Table.ts
```diff
@@ -38,6 +38,8 @@ export interface TableColumn<V, R, SV, SR> {
   label?: string
   className?: string
   dropdown?: DropdownSection[]
+  grow?: boolean
+  fillOnAdjust?: boolean
```
kiaking, line 42: What does `fillOnAdjust` option do? 👀

brc-dd: It needs better name 😅. It basically the column that expands itself to occupy rest of the space when any col is adjusted. It should be non-resizable -- usually the last column.

kiaking: Ahhh, totally make sense.

> We can probably keep it to always be the last column without need of any extra prop.

I like this idea. I guess we'll end up defining this option almost 100% the time so I think we could start by removing this option and just make it default behavior 👀

PR #241, opened 2023-04-06
feat(button): add icon-mode prop

### LINKED ISSUE #235: [Button] add option to change only the icon color
It would be cool if we can make button similar to what GitHub has.

Screenshot 2023-04-05 at 12 54 39

Maybe we could add `:icon-mode` prop for this?

```ts
interface Props {
  iconMode?: Mode
}

type Mode =
  | 'neutral'
  | 'info'
  | 'success'
  | 'warning'
  | 'danger'
```

brc-dd:
fixes #235

### DIFF
--- a/lib/components/SButton.vue
+++ b/lib/components/SButton.vue
@@ -24,6 +24,7 @@ const props = defineProps<{
   type?: Type
   mode?: Mode
   icon?: any
+  iconMode?: Mode
   label?: string
   href?: string
   rounded?: boolean
@@ -69,7 +70,7 @@ function handleClick(): void {
     @click="handleClick"
   >
     <span class="content">
-      <span v-if="icon" class="icon"><SIcon :icon="icon" class="icon-svg" /></span>
+      <span v-if="icon" class="icon"><SIcon :icon="icon" class="icon-svg" :class="iconMode" /></span>
       <span v-if="label" class="label">{{ label }}</span>
     </span>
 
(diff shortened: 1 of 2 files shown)


docs/components/button.md
```diff
@@ -275,186 +289,234 @@ The component has several different styles based on its type and color combinati
 :root {
   --button-fill-neutral-border-color: transparent;
   --button-fill-neutral-text-color: var(--c-text-inverse-1);
+  --button-fill-neutral-icon-color: var(--c-text-1);
```
kiaking, line 292: Thanks for adding it, but, let's use `text` colors for icons too, since that's what usually Sefirot does. I should have named this `fg` (foreground) instead 😓 That's what it was meant. The color that used on foreground, (which is usually text)

kiaking (commented):
Thanks a lot! Added a comment 🙏

kiaking:
And sorry can we also include `label-mode` that changes only the label color of the button? To have this kind of style.

Screenshot 2023-04-07 at 10 14 17

This should be done by `<SButton :mode="mute" :label-mode="danger" />`.

BTW, we could have `content-mode` prop that would change both icon and label color but not BG at this pont... though that might be too much (we can simply define 2 props if there's such a use case). So, I think we wouldn't need it 🤔

docs/components/button.md
brc-dd: Yeah, I tried doing that first 😅. But the issue is the text color for `fill` mode behave differently than `outline`/`text` mode.

kiaking: Ahhh... yeah... true. Thanks! I guess there's no way around this then 😅

kiaking (approved):

PR #244, opened 2023-04-10
perf(table): use virtualization and reduce reactivity overhead

### LINKED ISSUE #243: [Table] Stop using `reactive` in `useTable`
We've got issue where when we have many records (over 500 records), the table becomes very slow. Seems like getting rid of `reactive` drastically reduce the render time.

So, let's stop using `reactive` in `useTable`. The reason why `useTable` wraps the options with `reactive` is to unwrap refs before passing it in to `<STable>` as props.

In order to not use `reactive`, we need to modify `Table` type to accept `MaybeRef`. And then modify `<STable>` related components to accept plain refs.

We still want `useTable` composable just for typings, but after this change, the function should just return its argument. It should look like this I guess.

```ts
export function useTable<
  O extends string,
  R extends Record<string, any>,
  SR extends Record<string, any>
>(
  options: Table<O, R, SR>
): Table<O, R, SR> {
  return options
}
```

brc-dd:
fixes #243

For testing do something like this in story:

```ts
for (let i = 0; i < 10000; ++i) {
  data.push({
    name: `Artwork ${i}`,
    link: 'https://example.com',
    status: 'Published',
    type: 'Photo',
    createdAt: '2022-10-10',
    tags: ['Info', 'News', 'Latest']
  })
}
```

And set `--table-max-height` to some fixed value (e.g. `65vh` or maybe something complex like `calc(100vh - 5rem)`)

TODO

- [x] fix table styles in docs

### DIFF
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -345,6 +345,7 @@ function getCell(key: string, index: number) {
     border-radius: 0;
   }
 
+  scrollbar-width: none;
   &::-webkit-scrollbar {
     display: none;
   }
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -352,13 +352,13 @@ function getCell(key: string, index: number) {
 
 .container.body {
   border-radius: 6px 6px var(--table-border-radius) var(--table-border-radius);
+  line-height: 0;
 
   .STable.has-footer & {
     border-radius: 0;
   }
 
   .block {
-    margin-bottom: -5px;
     max-height: var(--table-max-height, 100%);
     overflow-y: auto;
   }
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -10,7 +10,6 @@ import {
   watch
 } from 'vue'
 import { DynamicScroller, DynamicScrollerItem } from 'vue-virtual-scroller'
-import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
 import type { Table } from '../composables/Table'
 import SSpinner from './SSpinner.vue'
 import STableCell from './STableCell.vue'
@@ -29,17 +28,24 @@ const block = shallowRef<HTMLElement | null>(null)
 const row = shallowRef<HTMLElement | null>(null)
 
 const colToGrowAdjusted = ref(false)
+
 const colToGrow = computed(() => {
-  return colToGrowAdjusted.value
-    ? -1
-    : unref(props.options.orders).findIndex(
-      (key) => unref(props.options.columns)[key]?.grow
-    ) ?? -1
+  if (colToGrowAdjusted.value) {
+    return -1
+  }
+
+  return unref(props.options.orders).findIndex((key) => {
+    return unref(props.options.columns)[key]?.grow
+  }) ?? -1
 })
-const nameOfColToGrow = computed(
-  () => unref(props.options.orders)[colToGrow.value]
-)
-const cellOfColToGrow = computed(() => row.value?.children[colToGrow.value])
+
+const nameOfColToGrow = computed(() => {
+  return unref(props.options.orders)[colToGrow.value]
+})
+
+const cellOfColToGrow = computed(() => {
+  return row.value?.children[colToGrow.value]
+})
 
 let headLock = false
 let bodyLock = false
@@ -75,6 +81,14 @@ const showFooter = computed(() => {
   )
 })
 
+const showMissing = computed(() => {
+  return (
+    !unref(props.options.loading)
+    && unref(props.options.records)
+    && !unref(props.options.records)!.length
+  )
+})
+
 const classes = computed(() => ({
   'has-header': showHeader.value,
   'has-footer': showFooter.value,
@@ -86,6 +100,7 @@ const recordsWithSummary = computed(() => {
   const summary = unref(props.options.summary)
 
   const res = summary ? [...records, summary] : records
+
   res.forEach((record, index) => {
     record.__index = index
   })
@@ -172,6 +187,14 @@ function updateColWidth(key: string, value: string, triggeredByUser = false) {
   }
 }
 
+function isSummaryOrLastClass(index: number) {
+  if (isSummary(index)) {
+    return 'summary'
+  }
+
+  return lastRow(index) ? 'last' : ''
+}
+
 function isSummary(index: number) {
   return index === unref(props.options.records)?.length
 }
@@ -250,16 +273,7 @@ function getCell(key: string, index: number) {
                 :active="active"
                 :data-index="rIndex"
               >
-                <div
-                  class="row"
-                  :class="
-                    isSummary(rIndex)
-                      ? 'summary'
-                      : lastRow(rIndex)
-                        ? 'last'
-                        : ''
-                  "
-                >
+                <div class="row" :class="isSummaryOrLastClass(rIndex)">
                   <STableItem
                     v-for="key in options.orders"
                     :key="key"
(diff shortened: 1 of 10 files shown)


lib/components/STable.vue
```diff
@@ -9,6 +9,8 @@ import {
   toRefs,
   watch
 } from 'vue'
+import { DynamicScroller, DynamicScrollerItem } from 'vue-virtual-scroller'
+import 'vue-virtual-scroller/dist/vue-virtual-scroller.css'
```
brc-dd, line 13: I earlier added it to bootstrap.css but it's not being imported in VitePress so styles were breaking there 👀

lib/components/STable.vue
```diff
@@ -108,6 +112,10 @@ watch(() => records?.value, () => {
   bodyLock = false
 }, { flush: 'post' })
 
+useResizeObserver(block, ([entry]) => {
+  blockWidth.value = entry.contentRect.width
```
brc-dd, line 116: Earlier tried using useElementSize/useElementBounding, but those were increasing glitches (delay between head and body) in scrolling.

`blockWidth` can be entirely removed if we can somehow do this from CSS (make the body scroll horizontally). Sadly nothing along the lines of fit-content worked for me.

lib/components/STable.vue
```diff
@@ -175,6 +183,10 @@ function isSummary(index: number) {
   return index === records?.value?.length
 }
 
+function lastRow(index: number) {
+  return index === recordsWithSummary.value.length - 1
```
brc-dd, line 187: This too can be avoided if we can get the last row using CSS. From what I can tell, the order of rows isn't same in resultant HTML at least initially. But if the scroll is at the bottom, then the last row in HTML was actually the last. So might be possible using CSS only 👀

lib/components/STable.vue
```diff
@@ -338,20 +366,27 @@ function getCell(key: string, index: number) {
   .STable.has-footer & {
     border-radius: 0;
   }
+
+  .block {
+    margin-bottom: -5px;
+    max-height: var(--table-max-height, 65vh);
```
brc-dd, line 372: This new CSS variable is introduced. It's necessary to limit height in order for virtual-scroller to work.

lib/components/STable.vue
```diff
@@ -338,20 +366,27 @@ function getCell(key: string, index: number) {
   .STable.has-footer & {
     border-radius: 0;
   }
+
+  .block {
+    margin-bottom: -5px;
```
brc-dd, line 371: Not sure why this is needed. There is a consistent gap after the last row of ~5.09px. Can't understand where it's coming from 🥲.

lib/components/STable.vue
```diff
@@ -338,20 +366,27 @@ function getCell(key: string, index: number) {
...
+  }
 }
 
 .block {
   display: inline-block;
   min-width: 100%;
 }
 
-.row {
+:deep(.row) {
```
brc-dd, line 382: Made these deep because row are rendered both outside/inside slots. `.row, :slotted(.row)` would also work.

tests/components/STable.spec.ts
```diff
@@ -24,29 +24,31 @@ describe('components/STable', () => {
...
-      },
-      records: [
-        { name: 'Item 1', amount: 10 },
-        { name: 'Item 2', amount: 90 }
-      ],
-      summary: {
-        name: 'Total', amount: 100
-      }
-    })
+  // FIXME: Does not work because of virtualization
```
brc-dd, line 27: Disabled this test for now. Summary row is rendered properly, but will need to update the logic to handle virtual scrolling.

kiaking:
@brc-dd 
Thanks a lot! Haven't gone through yet but 1 quick comment first. I still think we should eliminate `shallowReactive`.

Because `shallowReactive` will not unwrap refs. Although it should work fine since we're using `toRefs` in `<STable>` and treating everything as refs, however now `Table` type is technically wrong 😅 (many property should have `MaybeRef`).

---
@ryo-gk @NozomuIkuta 
Could you test this out to see if the issue we're having is solved?

Also, [we now have `max-height`](https://github.com/globalbrain/sefirot/pull/244/files#diff-45cde3b0098d79eb51ac6344afe5cbb422414cac2caaefb57fb2abef3e0cccadR372). I would like to know how would this affect on our ordinary usage where we have for example only 1 row of item, or what happens when scrolling entire window?

NozomuIkuta:
@brc-dd 

Thank you for the PR!

I instantly checked `STable` component in Histoire with 10000 records as you described in PR message, found that pagination UI is not working and all records are shown in one page.

Would you keep the existing features working in addition to enabling virtual scrolling in case there are lots of records in one page?

brc-dd:
> I instantly checked STable component in Histoire with 10000 records as you described in PR message, found that pagination UI is not working and all records are shown in one page.

Actually pagination was never enabled in Histoire. For pagination we need to add those onNext/onPrev functions in story that slice the records.

kiaking:
Yeah pagination has nothing to do with this change 👀 User should handle pagination. Not STable.

kiaking:
@brc-dd
We've tested the table and it's working really good in most of the cases. We found in some case, `row.last` is applied at wrong record. Not sure why (it only happens on few table and not all).

So, after you finish up getting rid of `toRefs` thing, we'll take over and try to solve all issues 👍

Also, probably we'll set `max-height` default to `100%` to keep current behavior. However, you can leave these tiny stuff for me.

brc-dd:
> We found in some case, row.last is applied at wrong record. Not sure why (it only happens on few table and not all).

Yeah that's because the underlying elements are reused during scroll. Will see if I can fix that too.

brc-dd:
> Also, probably we'll set max-height default to 100% to keep current behavior.

This might not work as the table height needs to be fixed in order to make it scrollable 👀

lib/components/STable.vue
```diff
@@ -225,62 +224,81 @@ function getCell(key: string, index: number) {
...
-            >
-              <STableItem
-                v-for="key in orders"
-                :key="key"
-                :name="key"
-                :class-name="columns[key].className"
-                :width="colWidths[key]"
+          <DynamicScroller
+            :items="recordsWithSummary"
+            :min-item-size="40"
```
brc-dd, line 236: might need to change this value too 👀

kiaking:
> > Also, probably we'll set max-height default to 100% to keep current behavior.
> 
> This might not work as the table height needs to be fixed in order to make it scrollable 👀

Yeah that's OK. Currently, most of the table we use don't have inline scrolling and we rather just scroll the body instead. Which will not benefit from virtual listing, but that's OK for small size records.

Inline scroll has downside where you must hover over table in order to scroll. Which might not be ideal for some use cases.

Where we need virtual listing, we can set max height explicitly.

Regarding this inline scroll behavior, I do want to add another feature. It's a bit complicated to explain in text so maybe I'll make video explaining it, or call you 👀

lib/components/STable.vue
kiaking: Ah. What does this do? It sets min height...?

brc-dd: Minimum height of a row.

kiaking: Gotcha. I'll set this to match the height of the body when spinner is shown 👍 (when `loading: true`.

kiaking: Ah yeah so that's kinda inevitable because sometimes we only have 1 record 😅 (depending on applied filters and such).

And we always get layout shift when paginating due to showing spinner 👀

lib/components/STableItem.vue
brc-dd, line 1: These changes were necessary to a fix a bug introduced in last PR. If `grow` isn't set on any property, then last column will collapsed weirdly. (Try scrolling this table in mobile: https://sefirot.globalbrains.com/components/table)

lib/components/STable.vue
kiaking: I'll add this to VitePress 👍 In case if it doesn't work, I'll just place it here 😅

kiaking (approved):
We've finished testing and this is working really well! This is clearly the one of the best thing happened to Sefirot. Thank you so much @brc-dd!

PR #251, opened 2023-04-14
feat(table): add `TableCellState` type

### LINKED ISSUE #248: [Table] Add `TableCellState` type
require https://github.com/globalbrain/sefirot/issues/247

Add `TableCellState` type to display `<SState>` in the table cell.

Screenshot 2023-04-13 at 11 31 05

## API

```ts
interface TableCellState extends TableCellBase {
  type: 'state'
  label?: string
  mode?: Mode // `<SState>` modes.
}
```

**NOTE:** We don't need callback sygnature. I think we can make it deprecated for other fields that currently have since we now can make `TableCell` it self a callback in `columns` option.

```ts
// If you need to do this.
{
  cell: {
    type: 'state',
    value: (_, record) => record.someThing
  }
}

// Do this instead.
{
  cell: (_, record) => ({
    type: 'state',
    value: record.someThing
  })
}
```

cuebit:
resolves #248

Screenshot 2023-04-14 at 17 16 05

### DIFF
--- a/lib/components/STableCell.vue
+++ b/lib/components/STableCell.vue
@@ -7,6 +7,7 @@ import STableCellDay from './STableCellDay.vue'
 import STableCellEmpty from './STableCellEmpty.vue'
 import STableCellPill from './STableCellPill.vue'
 import STableCellPills from './STableCellPills.vue'
+import STableCellState from './STableCellState.vue'
 import STableCellText from './STableCellText.vue'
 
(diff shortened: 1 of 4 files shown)


brc-dd (approved):

ryo-gk (approved):

kiaking (approved):
Niiice! Perfect. Thank you so much 🙌

PR #252, opened 2023-04-17
test(SState): missing required prop

cuebit:
Removes multiple warnings, for example:

```
stderr | tests/components/SState.spec.ts > components/SState > tag is set to `span` by default
[Vue warn]: Missing required prop: "label" 
  at <SState ref="VTU_COMPONENT" > 
  at <VTUROOT>
```


kiaking (approved):
Ah, thanks!

PR #260, opened 2023-04-20
feat(dropdown): add disabled options

ryo-gk:
This PR adds `disabled` options to `DropdownSectionFilterOptionBase` so that it makes it possible to show only enabled options on the dropdown.
You can write like the below example, and then, only 'Draft' and 'Published' are shown on the dropdown.

```vue
<script setup>
import { ref } from 'vue'
import SDropdown from 'sefirot/components/SDropdown.vue'
import { createDropdown } from 'sefirot/composables/Dropdown'

const selected = ref('Draft')
const sections = createDropdown([
  {
    type: 'filter',
    selected,
    options: [
      { label: 'Draft', value: 'Draft' },
      { label: 'Published', value: 'Published' },
      { label: 'Archived', value: 'Archived', disabled: true }
    ],
    onClick: () => {}
  }
])
</script>

<template>
  <SDropdown :sections="sections" />
</template>
```


kiaking (approved):
Nice! I've moved the test to `SDropdown.spec.ts` since this is part of the `SDropdown` component, and I think it makes more sense to group these kind of components together.

Issue #255, opened 2023-04-20
A good way to provide more info to users on disabled actions


kiaking:
This issue is wrapper issue that consists of many different tasks so that we can solve a particular problem.

## What we wanna do

Many apps in GB has quite complex authorizations to perform some actions like updating objects. When user is not permitted to perform an action, we tend to make a button disabled like this.

<img width="638" alt="Screenshot 2023-04-20 at 9 45 06" src="https://user-images.githubusercontent.com/3753672/233229542-08d7089c-0480-4d34-8c69-bc0e0ed318a3.png">

The above example shows that "Update tags" button is not available.

The issue is, the reason why it's disabled is not clear to users. Sometimes, the button is disabled because the user is lacking the authorization permission. But sometimes, maybe the button is disabled because the object state is not right (e.g. If an object has status "Completed", then user should not be able to update it).

Simply disabling or hiding button make users wonder why? And we would like to have some way to nicely handle this situation.

## Clickable tooltip

One thing that comes to our mind very quickly is to use tooltips. However, showing tooltips on hover is not a good idea in terms of touch devices. It's almost impossible to show tooltips on button.

So, instead of showing the tooltip on hover, how about if we show tooltip when user actually click the disabled button?

<img width="552" alt="Screenshot 2023-04-20 at 10 28 22" src="https://user-images.githubusercontent.com/3753672/233234337-edb959f2-8d34-45fd-859e-212e95961d65.png">

This way we can add description to actions that needs explanation to understand why something is permitted or not.

Sidenote: We should also consider doing the same for `:info` option on input components (make it clickable tooltip rather than activate on hover) 👉 https://github.com/globalbrain/sefirot/issues/259

## Todos

Related issues in order to make this work.

- [x] https://github.com/globalbrain/sefirot/issues/256
- [x] https://github.com/globalbrain/sefirot/issues/257
- [x] https://github.com/globalbrain/sefirot/issues/258

PR #261, opened 2023-04-21
feat(SButton): emit disabled-click (#255)

### LINKED ISSUE #257: [Button] Emit `@disabled-click` event
- ref #255

Currently, `SButton` will not emit `@click` event when it's disabled, which is good in most of the cases. By adding special click event that emits only when button is disabled will enable us to hook into such event and show something to users. Refer #255 for motivation behind this.

```ts
interface Events {
  // Only fires when user clicks on disabled button.
  (e: 'disabled-click'): void
}
```

cuebit:
resolves #257

### DIFF
--- a/lib/components/SButton.vue
+++ b/lib/components/SButton.vue
@@ -36,6 +36,7 @@ const props = defineProps<{
 
 const emit = defineEmits<{
   (e: 'click'): void
+  (e: 'disabled-click'): void
 }>()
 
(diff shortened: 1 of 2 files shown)


kiaking (approved):
Nice! Thanks 🙌 I've added the docs 💪

cuebit:
Ah, completely overlooked. Thanks for adding the docs!

PR #262, opened 2023-04-24
feat(tooltip): allow using focus as trigger and specifying timeout

### LINKED ISSUE #256: [Tooltip] Add "Click to activate" option
- ref #255 

Add option to show tooltip on click instead of hover. Maybe something like this?

```ts
interface Props {
  // Defaults to 'hover'.
  when?: 'click' | 'hover'

  // In ms. Close the tooltip after this time has passed.
  // Only relevant when `:when` is set to `'click'`.
  // Infinite if it's `undefined`.
  timeout?: number
}
```

```html
<STooltip when="click">
  ...
</STooltip>
```

brc-dd:
fixes #256

TODO:
- [x] tests
- [x] docs


brc-dd:
Just noticed a thing, that SMarkdown thing is not working in any of my browsers except Chrome (desktop). It shows blank on all mobile browsers and firefox/safari. Happening with deployed docs too, so related with this PR IG.

brc-dd:
Will work with buttons without the need of disabled-click/click event:

image

cuebit:
Can confirm this works on disabled buttons. We can revert the `@disabled-click` event after this is merged.

FYI, can we declare top-level functions using the function keyword? This will be auto linted when the eslint-config is updated in any case 👍

brc-dd:
> FYI, can we declare top-level functions using the function keyword?

The `onMouseEnter/Leave`, etc.? Yeah, I'll update them.

kiaking:
> Can confirm this works on disabled buttons. We can revert the @disabled-click event after this is merged.

I think we can keep this event. Still could useful when we want to hook into disabled click event 👀

kiaking:
I've also tested and it's working smooth as hell. VERY NICE JOB.

I like the props names too! Let me know when you're done working on. I'll like to modify just a bit on styling (not really related to this PR but like line-height and such).

Also, when the `:trigger` is set to `focus` or `both`, would it make sense to make the cursor `pointer`...? 🤔 I'm not 100% sure.

brc-dd:
> Also, when the :trigger is set to focus or both, would it make sense to make the cursor pointer...?

Yeah we can do that, but if we are just wrapping the buttons, they'll already have that style.

kiaking:
>> Also, when the :trigger is set to focus or both, would it make sense to make the cursor pointer...?
>
> Yeah we can do that, but if we are just wrapping the buttons, they'll already have that style.

Good point! Though, tooltip might be used for other elements too so 👀

brc-dd:
Added it!

vitest.config.ts
```diff
@@ -13,7 +13,7 @@ export default defineConfig({
 
   test: {
     globals: true,
-    environment: 'jsdom',
+    environment: 'happy-dom',
```
brc-dd, line 16: Has to switch back to `happy-dom` because `jsdom` throws errors on `:focus-within`. `happy-dom` too doesn't support that selector, but it silently ignores 😅

kiaking:
Thanks! I'll do the modify now and merge 💪

vitest.config.ts
kiaking: 👍

kiaking:
I've updated the tooltip styles and format a bit. I'll merge this and create another PR for Tooltip. Want to adjust the bg color and such but need to fix some global variables first.

PR #267, opened 2023-04-25
chore(story): fix router symbol not found warnings

### LINKED ISSUE #265: `injection "Symbol(router)" not found` everywhere in Histoire
Kinda annoying so would be cool if we can get rid of this someway 🤔 Maybe we need to add `:setup-app="setupRouter"` everywhere?

brc-dd:
fixes #265


kiaking (approved):
OK that was easy 😅  Thanks!

PR #268, opened 2023-04-25
fix(grid): instantiate MutationObserver on client only (#263)

### LINKED ISSUE #263: [Grid] options not optional and doesn't work in SSR
`useGrid` needs fixes.

## `options` argument is not optional

All fields in `options` object is optional but the argument it self is not.

```ts
function useGrid(options: UseGridOptions): Grid

// Should be
function useGrid(options: UseGridOptions = {}): Grid
```

## Doesn't work in SSR

Because `MutationObserver` doesn't exist on server. For example, it throws on VitePress build. Probably we should do something like this?

```ts
export function useGrid(options: UseGridOptions): Grid {
  // Assign this on `onMounted` hook.
  let observer: MutationObserver | null = null

  watchEffect(() => {
    observer?.disconnect()

    if (container.value) {
      adjustSpacer()
      observer?.observe(container.value, { childList: true })
    }
  })

  onMounted(() => {
    // Here.
    observer = new MutationObserver((_, observer) => {
      observer.disconnect()
      adjustSpacer()
      observer.observe(container.value!, { childList: true })
    })

    ...
  })

  ...
}
```

cuebit:
resolves #263

### DIFF
--- a/lib/composables/Grid.ts
+++ b/lib/composables/Grid.ts
@@ -20,25 +20,25 @@ export function useGrid(options: UseGridOptions): Grid {
   const spacerTag = options.tag ?? 'div'
   const type = options.type ?? 'fit'
 
-  const observer = new MutationObserver((_, observer) => {
-    observer.disconnect()
-
-    adjustSpacer()
-
-    observer.observe(container.value!, { childList: true })
-  })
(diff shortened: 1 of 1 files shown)


kiaking (approved):
Thanks a bunch!

PR #269, opened 2023-04-26
fix: error on set placeholder partially

ryo-gk:
This PR fixes the issue that it causes the error when we set the placeholder to `SInputYMD` and `SInputHMS` partially.

### DIFF
--- a/lib/components/SInputHMS.vue
+++ b/lib/components/SInputHMS.vue
@@ -63,9 +63,9 @@ const padValue = computed(() => {
 
 const padPlaceholder = computed(() => {
   return {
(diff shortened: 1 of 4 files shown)


kiaking (commented):
We need to update the `Placeholder` types as well since they are not nullable 👀  It should be:

```ts
export interface Placeholder {
  hour?: string
  minute?: string
  second?: string
}
```

kiaking:
Same goes for `SInputYMD` too 👀

ryo-gk:
Ah, I mistook the Placeholder in SInputYMD for Value and thought it was okay. 😓 
I fixed it.

PR #274, opened 2023-05-10
chore: bump deps

### LINKED ISSUE #273: [InputDate] `v-calendar` version is stale
While Sefirot depends on `v-calendar@v3.0.0-alpha.8`, its upstream version at the moment is `v3.0.3`.
It's worth checking if we can upgrade the dependency.

### LINKED ISSUE #276: [Button] label is not centered when used inside `SSheetFooterAction`
when `SButton`'s `block` prop is enabled and the button is used inside `SSheetFooterAction`, the button label comes to look like below: 

Screenshot 2023-05-10 at 17 14 35

As far as I checked, the text element doesn't extend (or "flex-grow") in the button, which is in turn extended to [min-width of `SSheetFooterAction`](https://github.com/globalbrain/sefirot/blob/95d2d34f5aa6e725fec11e43d88e7183f2d0901c/lib/components/SSheetFooterAction.vue#L31-L35).

Workaround at the moment is to define styles like below where `SSheetFooterAction` is used.

```vue
<style scoped lang="postcss">
:deep(.SSheetFooterAction .SButton .content) {
  flex-grow: 1
}
</style>
```

By the way, if `block` prop is disabled, then the button comes to not extend to min-width of `SSheetFooterAction` like below:

Screenshot 2023-05-10 at 17 23 47

brc-dd:
closes #273
closes #276


lib/components/SButton.vue
```diff
@@ -705,6 +705,7 @@ function handleClick(): void {
 
 .SButton.block {
   display: flex;
+  justify-content: center;
```
brc-dd, line 708: [SSheet had this misaligned.](https://story.sefirot.globalbrains.com/story/stories-components-ssheet-01-playground-story-vue?variantId=_default) So fixed it too.

kiaking: Thanks! 🙌

kiaking (approved):

PR #277, opened 2023-05-22
fix: error on accessing nullable variables

ryo-gk:
Fix the error which occurs when `container` element is null.

### DIFF
--- a/lib/composables/Flyout.ts
+++ b/lib/composables/Flyout.ts
@@ -19,13 +19,13 @@ export function useFlyout(container?: Ref<any>) {
   }
 
   function closeOnClickOutside(event: any) {
-    if (!el.value.contains(event.target) && isVisible(el.value)) {
+    if (!el.value?.contains(event.target) && isVisible(el.value)) {
       isOpen.value = false
     }
   }
 
   function isVisible(el: any) {
-    return !!el && !!(el.offsetWidth || el.offsetHeight || el.getClientRects().length)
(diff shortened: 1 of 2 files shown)


kiaking (commented):
Looks reasonable to me.

@brc-dd Could you take a look? 👀

lib/composables/Flyout.ts
```diff
@@ -19,13 +19,13 @@ export function useFlyout(container?: Ref<any>) {
...
   function closeOnClickOutside(event: any) {
-    if (!el.value.contains(event.target) && isVisible(el.value)) {
+    if (!el.value?.contains(event.target) && isVisible(el.value)) {
       isOpen.value = false
     }
   }
 
   function isVisible(el: any) {
-    return !!el && !!(el.offsetWidth || el.offsetHeight || el.getClientRects().length)
+    return !!el && !!(el?.offsetWidth || el?.offsetHeight || el.getClientRects().length)
```
brc-dd, line 28: these might not be necessary since there is already a `!!el` check.

brc-dd (approved):

brc-dd:
LGTM, but why aren't these handled by TS in the first place? (I see that everything in the flyout composable is `any`.)

kiaking:
> LGTM, but why aren't these handled by TS in the first place? (I see that everything in the flyout composable is any.)

I'm pretty sure because we were too laze 😅

lib/composables/Flyout.ts
ryo-gk: Thanks, you right!
I fixed it.

PR #278, opened 2023-05-22
feat(card): add `<SCard>` component

kiaking:
Add a new `<SCard>` component.

[See Docs](https://deploy-preview-278--sefirot-docs.netlify.app/components/card) for detailed usage.

This does not cover all use cases or GB's apps, but I think this is a starting point.

One thing I had to do some hacky stuff on `<SCardHeaderActionCollapse>` where it sets class name to `<SCard>` by looking up the DOM, but I couldn't find a better way to do it 🤔 

I wanted to make the usage consistent on how we place all nested components without adding props (for example `collapsable`) to `<SCard>` component.

---

Screenshot 2023-05-22 at 17 45 53


ryo-gk:
It looks good to me 🙌 
I think it might be nice to have a `tooltip` prop on `HeaderAction` and `FooterAction` since it can be used with a `disabled` prop in some cases, but is it unnecessary for now?

kiaking:
@ryo-gk Ahhh good point. Yes we should have it. If everybody reviews this PR soon enough, probably merge first and then add as a new feature but maybe I could just include in this PR too. Lets see 👀

brc-dd (approved):
LGTM.

> One thing I had to do some hacky stuff on `<SCardHeaderActionCollapse>` where it sets class name to `<SCard>` by looking up the DOM, but I couldn't find a better way to do it 🤔

You can use provide/inject for that 👀

kiaking:
@brc-dd 

> You can use provide/inject for that 👀

Good point! I've thought about it too, but wasn't sure if it works with slot 🤔... I'll try before merging this. I think it should look better that way rather than DOM queries 😅

kiaking:
OK the provide/inject indeed worked with slot 😳 So, I've updated. Also added `:tooltip` option to actions as well 👍

PR #280, opened 2023-05-23
feat(dropdown): show radio button when selected is not array

kiaking:
Show radio button instead of checkbox on dropdown filter items when `selected` is not an array. Because when selectable item is stored as primitive value, that means user can only select an item a time.

Also I've refined the design as well.

Screenshot 2023-05-23 at 12 36 50


ryo-gk (approved):
LGTM

brc-dd (approved):

brc-dd (approved):

PR #282, opened 2023-05-23
feat(card): add special support when used within modal (#281)

### LINKED ISSUE #281: [InputDropdown] Enable passing in an object as `:value`
Currently, in order to use `<SInputDropdown>` component, we must have all `options` upfront in order to display which item is selected.

This is because currently the `:value` only accepts primitives (`string`, `number`, `boolean`, or `null`).

It would be very nice if we can pass in an option object as value and the component could use the object directly as a selected item in the input box before `:options` are set.

This is required when we want to fetch the option asynchronously, for example fetch all option only when user opens the dropdown menu.

The value should accept the same object as option.

```vue
<script setup lang="ts">
// Define value with object.
const value = ref({ label: 'John Doe', value: 1 })

// or array of ref for multi select
const value = ref([{ label: 'John Doe', value: 1 }])

const options = getOptions() // async call
</script>

<template>
  <SInputDropdown
    :options="options"
    v-model="value"
  />
</template>
```

When object is passed, the component should use it to display the value in input, and also use `value` property to determine which option is selected inside dropdown filter list.

kiaking:
- close #281

I went with the Idea 1 approach. I think this is much easier to consume and also, easier to add transition and box shadow.

[See it in action at histoire](https://deploy-preview-282--sefirot-story.netlify.app/story/stories-components-scard-02-within-modal-story-vue?variantId=_default).

The code would look like this.

```html
<SModal :open="isOpen">
  <SCard size="small">
    ...
  </SCard>
</SModal>
```

Screenshot 2023-05-23 at 13 45 22

### DIFF
--- a/lib/components/SCard.vue
+++ b/lib/components/SCard.vue
@@ -33,7 +33,7 @@ const { isCollapsed } = provideCardState()
 .SModal > .SCard {
   margin: 12px 12px 128px;
   box-shadow: var(--shadow-depth-3);
-  transition: opacity 0.5s, transform 0.5s;
+  transition: opacity 0.25s, transform 0.25s;
 
   @media (min-width: 512px) {
     margin: 24px 24px 128px;
--- a/lib/components/SCard.vue
+++ b/lib/components/SCard.vue
@@ -1,11 +1,17 @@
 <script setup lang="ts">
 import { provideCardState } from '../composables/Card'
 
+export type Size = 'small' | 'medium'
+
+defineProps<{
+  size?: Size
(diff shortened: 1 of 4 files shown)


brc-dd (approved):
LGTM 🚀 Transition seems bit slow though 😅

kiaking:
> Transition seems bit slow though

Hmmm maybe you're right. We might shall bump it up to 0.25s 🤔

kiaking:
OK I've updated to 0.25s 👍

PR #286, opened 2023-05-26
feat(num): add `abbreviate` utility

### LINKED ISSUE #285: [Support] Add `Num.abbreviate` funciton
We use this internally in GB's app and good have it in sefirot. We should add test and such. This method do things like:

```ts
Num.abbriviate(1_000, 0)     // '1K'
Num.abbriviate(1_250_000, 2) // '1.25M'
```

```ts
/**
 * Convert given number to abbriviated format such as `1K` or `1M`.
 */
export function abbreviate(num: number, precision = 0): string {
  const K = 1000
  const M = 1000000
  const B = 1000000000

  if (num >= K && num < M) {
    return `${(num / K).toFixed(precision)}K`
  }

  if (num >= M && num < B) {
    return `${(num / M).toFixed(precision)}M`
  }

  if (num >= B) {
    return `${(num / B).toFixed(precision)}B`
  }

  return String(num)
}
```

brc-dd:
fixes #285

### DIFF
--- a/docs/support/num.md
+++ b/docs/support/num.md
@@ -15,3 +15,16 @@ import { format } from '@globalbrain/sefirot/lib/support/Num'
 
 format(100000) // <- '100,000'
 ```
+
+## `abbreviate`
+
+```ts
+function abbreviate(value: number, precision = 0): string
+```
+
+```ts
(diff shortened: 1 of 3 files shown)


kiaking (approved):
P E R F E C T 🎉

PR #288, opened 2023-05-26
fix(tooltip): positioning breaking on some edge cases

kiaking:
When the tooltip overflows window width, seems like it isn't considering scroll bar width and causing windows to display temporally horizontal scroll bar.

image

With this, fix I've changed `windoe.outerWidth` to `document.body.clientWidth` to remove scroll bar width.

Also including `SCREEN_PADDING` (16px) when checking the tip position. When positioning tooltip we set 16px padding from the edge of the window. So, the tooltip must not exceed this 16px padding area. Which means the edge of the window should be `document.body.clientWidth - SCREEN_PADDING`.

I don't have windows so I can't directly confirm this but I'm asking to our team member for confirmation.

### DIFF
--- a/lib/composables/Tooltip.ts
+++ b/lib/composables/Tooltip.ts
@@ -67,9 +67,9 @@ export function useTooltip(
     const contentRightX = contentRect.x + contentRect.width
     const tipRightX = tipRect.x + tipRect.width
 
-    if (tipRect.x < 0) {
+    if (tipRect.x < SCREEN_PADDING) {
       adjustLeftPosition(contentRect.x)
-    } else if (tipRightX > window.outerWidth) {
+    } else if (tipRightX > (document.body.clientWidth - SCREEN_PADDING)) {
       adjustRightPosition(contentRightX)
     }
 
@@ -85,7 +85,7 @@ export function useTooltip(
   function adjustRightPosition(contentRightX: number): void {
     tip.value!.style.left = 'auto'
     tip.value!.style.right = '0'
-    setTransform((window.outerWidth - contentRightX) - SCREEN_PADDING)
+    setTransform((document.body.clientWidth - contentRightX) - SCREEN_PADDING)
   }
 
   function resetPosition(): void {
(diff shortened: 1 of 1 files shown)


brc-dd (approved):

brc-dd:
> I don't have windows so I can't directly confirm this but I'm asking to our team member for confirmation.

temporarily enable it:

image

kiaking:
@brc-dd Yeah tried that but seems like that option would not enable scroll bar when the window is not scrollable... 🤔

Screenshot 2023-05-26 at 19 47 36

The scrollbar will show when the screen is scrollable.

Screenshot 2023-05-26 at 19 47 49

I don't think this was the behavior before. Maybe my machine is having some kind of trouble, or Mac update changed something...?

brc-dd:
> The scrollbar will show when the screen is scrollable.

That's the same behavior on Windows as well. Scrollbar doesn't appear there unless there is anything to scroll (or you've explicitly set `overflow:scroll`). (That's why we get a lot of complaints on VitePress that changing page, when one has lesser content, shifts things 😅)

kiaking:
Oh... yeah that's true... then I'm not really sure anymore why we are getting this issue 🤔 ... hmmm... Well, I'll wait for the windows check and see how it goes 😇

brc-dd:
Update `SCREEN_PADDING` to something like 20px? (it's better to leave some space from boundary anyway)

kiaking:
Hmmmm good point. @cuebit Could you try setting `SCREEN_PADDING` to `24px`. I believe scroll bar width is wider than 16px, I guess it's more like 18–20.

However, we should check for the view port in the end. 24px gap is too wide for mobile so probably for mobile we set to like 8px, and set 24px when viewport is bigger than 512px 🤔 

But first, need to confirm 24px gap will solve the issue 😳

kiaking:
OK our IT Corp team confirmed that setting it to 24px did solve the issue. So... maybe this is it 🤔

kiaking:
Aaaand updated the screen padding handling

PR #292, opened 2023-05-30
feat(table): add `onClick` option to `TableCellAvatar`

### LINKED ISSUE #291: [Table] Add `onClick` property for Avatar type cell
While text type cell accepts `onClick` property, avatar type doesn't.
We have UI design where user clicks text part of avatar cell and open a modal to edit record of the row.

NozomuIkuta:
Resolves #291.

I'm not sure if we can use `anonymous.png` and where to store it.


kiaking:
Ahh good point. I think this image is for example only so let's use GitHub avatar link so that we don't have to store anything in this repo 🤔  Everything looks good so I'll just gonna fix it and merge this 🙌

kiaking:
OK I've adjusted the format a bit and removed the image for now and reverted the story change, since I felt a bit awkward having exact same anonymous user avatar and the name was Artwork so 😅  We could always create a new story with avatar type focused example though 👍

PR #294, opened 2023-05-31
feat(link): add `external` prop to `SLink` (#293)

### LINKED ISSUE #293: [Link] same-origin link is always opened in the same browser tab
`SLink` set `target="_blank"` when value of `href` attribute starts with a string like external (e.g., `http:`).
We need to set `target="_blank"` for internal link in some case.

So, let's add `external` prop by which to override the result of regular expression matching like [`NuxtLink` does](https://nuxt.com/docs/api/components/nuxt-link).

`NuxtLink` has much flexibility, but we just need 3 use cases of them:

**Internal link in the same browser tab**

`<SLink href="/path/to/page">` 👉 `<a href="/path/to/page">`

**Internal link in another browser tab**

`<SLink href="/path/to/page" external>` 👉 `<a href="/path/to/page/" target="_blank" rel="noopener noreferrer">`

**External link in another browser tab (regardless of `external` prop value)**

`<SLink href="https://example.com">` 👉 `<a href="https://example.com" target="_blank" rel="noopener noreferrer">`
`<SLink href="https://example.com" external>` 👉 `<a href="https://example.com" target="_blank" rel="noopener noreferrer">`

NozomuIkuta:
Resolves #293

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -64,6 +64,7 @@ function sidebar(): DefaultTheme.SidebarItem[] {
         { text: 'SInputSwitch', link: '/components/input-switch' },
         { text: 'SInputTextarea', link: '/components/input-textarea' },
         { text: 'SInputYMD', link: '/components/input-ymd' },
+        { text: 'SLink', link: '/components/link' },
         { text: 'SPill', link: '/components/pill' },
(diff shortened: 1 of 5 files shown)


kiaking (approved):
I have:

- Fix `:external` default value issue (always false).
- Adjust story.
- Add test.
- Add docs.

PR #297, opened 2023-06-06
fix(table): No background color is applied when pills is empty array

ryo-gk:
This PR fixes that the background color of the `PillsCell` is different from other cells when the pills' value is empty array.
This is because the cell height is 0 when the pills are empty.


kiaking (approved):
Nice catch. Thanks!

PR #298, opened 2023-06-06
feat(table): add cell type `number` (#296)

### LINKED ISSUE #296: [Table] add "Number" type cell
Add `TableCellNumber` type. This type is different from `TableCellText` in terms of:

- It may have format option (`1000` -> `1,000`).
- It makes font mono space to align with all other records.
  - To do this, set css `font-feature-settings: "tnum"`. This will make all numbers have mono space width. This is feature of Inter font.

Other than above, it should extend all other options from `TableCellText`.

kiaking:
- close #296

Add a cell for number type. Although there is a lot of code duplication with the 'text' cell, I'm leaving it as is for now because we might introduce more number-specific features in the future, and I want to maintain flexibility.

Also, I've largely refactored the docs and separated it into multiple sections.

Screenshot 2023-06-06 at 17 59 59

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -60,7 +60,7 @@ const options = useTable({
 
 ## Usage
 
-Import `<STable>` component and `useTable` composable. Create table object with `useTable` composable and pass it to the `:options` props of the `STable` component. Here is the relatively simple table definitions.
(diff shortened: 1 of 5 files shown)


brc-dd (approved):
Code looks good to me. Still going through the docs.

brc-dd (approved):

PR #305, opened 2023-06-09
feat(table): add show option

### LINKED ISSUE #304: [Table] Add option to hide specific column
## Why?

There is a case where we would like to control the visibility of columns dynamically. For example, show some column only when some option is selected.

I thought we could do that by computing `orders` array.

```ts
const table = useTable({
  orders: computed(() => {
    const o = ['name', 'age']
    
    if (showEmail) {
      o.push('email')
    } 
  }),
  ...
})
```

However, we found out that this is going to break typings between `records` and `columns`. And also kinda verbose when we have many columns.

## Proposal

Instead of manipulating array contents, how about we always declare all possible options up front and add option to hide some specific column?

So in the above case, we would do something like this.

```ts
const table = useTable({
  // Declare all options.
  orders: computed(() => {
    const o = ['name', 'age', 'email']
  }),

  // Control the visibility as a column option.
  columns: computed(() => ({
    name: { ... },
    age: { ... },
    email: {
      label: 'Email',
      show: showEmail
    }
  }))
  ...
})
```

This way, all of the types are always preserved and looks very simple 👀

brc-dd:
fixes #304

### DIFF
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -27,6 +27,14 @@ const body = shallowRef<HTMLElement | null>(null)
 const block = shallowRef<HTMLElement | null>(null)
 const row = shallowRef<HTMLElement | null>(null)
 
+const ordersToShow = computed(() => {
+  return unref(props.options.orders).filter((key) => {
+    return unref(props.options.columns)[key]?.show !== false
+  })
+})
+
+watch(() => ordersToShow.value, handleResize)
+
 const colToGrowAdjusted = ref(false)
 
 const colToGrow = computed(() => {
@@ -34,13 +42,13 @@ const colToGrow = computed(() => {
     return -1
   }
 
-  return unref(props.options.orders).findIndex((key) => {
+  return ordersToShow.value.findIndex((key) => {
     return unref(props.options.columns)[key]?.grow
   }) ?? -1
 })
 
 const nameOfColToGrow = computed(() => {
-  return unref(props.options.orders)[colToGrow.value]
+  return ordersToShow.value[colToGrow.value]
 })
 
 const cellOfColToGrow = computed(() => {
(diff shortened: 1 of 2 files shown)


docs/components/table.md
```diff
@@ -313,6 +313,28 @@ const options = useTable({
...
+    age: { ... },
+    email: {
+      label: 'Email',
+      show: showEmail
+    }
+  }))
+  ...
+})
+```
+
```
kiaking, line 337: Maybe we should add `grow` option limitation here?

```
::: warning Limitation when using with `grow` option
When toggling the `show` option on a column with the `grow` option set to `true`, the auto-growing feature will be disabled. It is recommended to avoid using the `grow` option on columns that might be hidden.
:::
```

kiaking (commented):
Looking good! Shall we create a follow up issue on fixing the type issue with `computed` columns? 🤔 

Also, I'll make adjustment to story to use `menu` instead so that it looks more realistic as a GB app 👍

brc-dd:
> Also, I'll make adjustment to story to use menu instead so that it looks more realistic as a GB app 👍

Ah yeah, that'd be good. That button was just quick demo. Also, there are a bunch of errors on modifying refs in Histoire. (Similar errors happen while filtering things.)

> Shall we create a follow up issue on fixing the type issue with computed columns?

Yeah, working on that.

kiaking:
> Also, there are a bunch of errors on modifying refs in Histoire. (Similar errors happen while filtering things.)

Yeah this one right? `TypeError: 'set' on proxy: trap returned falsish for property 'hasFilters'`. I have no idea what this is but would be great to get rid of it 😅 

> > Shall we create a follow up issue on fixing the type issue with computed columns?
> 
> Yeah, working on that.

Wonderful 🙌

kiaking:
Adjusted the story. I think it got pretty. By the way, the grow feature works really good when toggling columns too. Very impressive.

Screenshot 2023-06-09 at 16 46 49

PR #309, opened 2023-06-20
feat(dropdown): allow passing props to component

### LINKED ISSUE #308: [Dropdown] Pass props to and listen to events from component
With current Dropdown section of type `component`, we can't pass props to the component and listen event from it (e.g. implementing `v-model` to it).

By supporting this feature, we can implement more dynamic and flexible component in dropdown like form items.

Note: it seems that `TableCellComponent` supports this feature.

brc-dd:
resolves #308, but not sure if it is required


kiaking (approved):
Thanks a lot!

PR #312, opened 2023-07-12
feat(card): add `large` size and `space` option to footer

kiaking:
Add `large` option to `<SCard>`. Only relevant when using inside `<SModal>`.

Also add `:space` option to `<SCradFooter>` so that it can have the same spacing with `<SCardBlock>` component. This spacing is needed to align with some of our app design.

Also, improved docs a bit.

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -184,7 +184,7 @@ Use `<SCardBlock>` to display generic block element. This component is usually u
 
 ### Spacing
 
-For combinience, `<SCardBlock>` component comes with `:space` props that lets you control the padding of the block. You may pass either `compact` or `wide` as a value.
+The `<SCardBlock>` component provides a convenient way to control the padding of the block using the `:space` prop. You can choose between two values: `compact` or `wide`.
 
 ```ts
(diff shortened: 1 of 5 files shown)


docs/components/card.md
```diff
@@ -239,19 +240,24 @@ export interface Tooltip {
...
-      label="Submit"
-      @click="onClick"
-    />
+  <SCardFooter space="compact">
+    ...
   </SCardFooter>
 </SCard>
 ```
+
+To ensure consistent spacing across the application, you should adjust the spacing based on the size of the `<SCard>` component. Consider using the `compact` class for card widths smaller than `640px` and the `wide` class for larger widths.
```
cuebit, line 261: Since these are props, not classes, I think _setting_ or _value_ makes sense here 😉 

```suggestion
To ensure consistent spacing across the application, you should adjust the `space` setting based on the size of the `<SCard>` component. Consider using the `compact` setting for card widths smaller than `640px` and the `wide` setting for larger widths.
```

cuebit (approved):

PR #313, opened 2023-07-12
feat(grid): add `<SGrid>` component

kiaking:
Add `<SGrid>` component.

This is a utility component that makes it a bit easy to define CSS Grid. Especially useful on apps that don't require responsive design. It just reduces a lot of boiler plate CSS and we quite use this a lot.

```html
<SGrid cols="4" gap="24">
  <SGridItem span="2">
    <div>...</div>
  </SGridItem>
  <SGridItem span="1">
    <div>...</div>
  </SGridItem>
</SGrid>
```

Screenshot 2023-07-12 at 14 53 29

### DIFF
--- a/docs/components/grid.md
+++ b/docs/components/grid.md
@@ -97,7 +97,7 @@ interface Props {
 
 ## Responsive design
 
-When you need to change the overall grid layout depending on the screen size, use plain CSS instead of using props. Those props are there for convenience, and you may always CSS to define complex layout structure.
+If you need to adjust the overall grid layout based on different screen sizes or other conditions, it is recommended to use CSS instead of props. The props are provided for convenience, but for more complex layout structures, CSS should be used.
 
 ```vue
 <template>
--- a/docs/components/grid.md
+++ b/docs/components/grid.md
@@ -1,0 +1,128 @@
+<script setup lang="ts">
+import SGrid from 'sefirot/components/SGrid.vue'
+import SGridItem from 'sefirot/components/SGridItem.vue'
+</script>
+
+# SGrid
+
+`<SGrid>` is a utility component to handle CSS grid layout.
+
+<Showcase
+  path="/components/SGrid.vue"
+  story="/stories-components-sgrid-01-playground-story-vue"
+>
+  <SGrid cols="4" gap="24">
+    <SGridItem>
+      <div class="h-64 rounded-6 bg-info" />
+    </SGridItem>
+    <SGridItem>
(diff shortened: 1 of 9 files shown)


docs/components/grid.md
```diff
@@ -0,0 +1,128 @@
...
+      <div>...</div>
+    </SGridItem>
+    <SGridItem span="1">
+      <div>...</div>
+    </SGridItem>
+  </SGrid>
+</template>
+```
+
+You have the flexibility to define the `:cols` and `:gap` props when using the `<SGrid>` component, allowing you to have full control over the grid layout.
```
cuebit, line 64: We don't label props with colons in the docs, should we remove here too to keep consistency?

```suggestion
You have the flexibility to define the `cols` and `gap` props when using the `<SGrid>` component, allowing you to have full control over the grid layout.
```

docs/components/grid.md
```diff
@@ -0,0 +1,128 @@
...
+```vue-html
+<SGrid cols="4" gap="48">
+  <SGridItem span="2">...</SGridItem>
+  <SGridItem span="2">...</SGridItem>
+</SGrid>
+```
+
+## Responsive design
+
+When you need to change the overall grid layout depending on the screen size, use plain CSS instead of using props. Those props are there for convenience, and you may always CSS to define complex layout structure.
```
cuebit, line 100: Maybe this reads better?

```suggestion
If you need to adjust the overall grid layout based on different screen sizes or other conditions, it is recommended to use CSS instead of props. The props are provided for convenience, but for more complex layout structures, CSS should be used.
```

docs/components/grid.md
cuebit: I've noticed there is a mixture of these existing in most recent PR's. Are you implying props should always be prefixed with a colon or is there some sort of context to why some props are prefixed and others are not throughout the docs?

kiaking: Ah, yeah... maybe. What do you think 🤔 Still not 100% sure. But maybe I'm thinking it could be easier to read if we follow these rules? 👀 

- `<SComponent>`
- `:prop`
- `@event`

cuebit (approved):

docs/components/grid.md
cuebit: > * `<SComponent>`
> * `:prop`
> * `@event`

Agree, this makes sense and gives connotations more clarity. Thanks for confirming!

kiaking: 👍

PR #320, opened 2023-07-20
fix(dropdown): reset white space wrapping

### LINKED ISSUE #319: [Table] Unexpected horizontal scroll on dropdown when item text is too long
We get this horizontal scroll effect.

Screenshot 2023-07-20 at 15 55 49

This only happens on `<STable>`. If we use `<SDropdown>` or `<SInputDropdown>`, the long text gets wrapped. It should behave the same for table too.

brc-dd:
closes #319


kiaking (approved):
Confirmed it's fixed! Thanks a bunch 🙌

PR #323, opened 2023-07-26
fix(grid): make styles overridable via CSS

### LINKED ISSUE #321: [Grid] Styles not able to override with CSS
We should be able to define `SGrid` component style by using `grid-template-column` and `gap` but this is not possible now due to inline styling.

When props are not defined, it should not define the inline style.

```ts
// SGrid

const styles = computed(() => {
  // Oh no... It should be undefined instead of having default values.
  return {
    gridTemplateColumns: `repeat(${props.cols ?? 1}, minmax(0, 1fr))`,
    gap: `${props.gap ?? 0}px`
  }
})
```

brc-dd:
fix #321 

Can be verified by something like this in `SGrid.01_Playground.story.vue`:

```html
<style scoped>
:deep(.SGrid) {
  grid-template-columns: repeat(2, minmax(0, 1fr));
  gap: 24px;
}
</style>
```

### DIFF
--- a/lib/components/SGrid.vue
+++ b/lib/components/SGrid.vue
@@ -6,16 +6,17 @@ const props = defineProps<{
   gap?: string | number
 }>()
 
-const styles = computed(() => {
-  return {
-    gridTemplateColumns: `repeat(${props.cols ?? 1}, minmax(0, 1fr))`,
-    gap: `${props.gap ?? 0}px`
-  }
+const gridTemplateColumns = computed(() => {
+  return `repeat(${props.cols ?? 1}, minmax(0, 1fr))`
+})
+
+const gap = computed(() => {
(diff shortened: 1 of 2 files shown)


kiaking (commented):
Confirmed! Thanks a bunch 🙌 

OK, now, how do we test this 🤔 ...

Screenshot 2023-07-26 at 12 40 08

brc-dd:
Huh, can't seem to make that work. Here is a solution: https://github.com/vuejs/test-utils/issues/1931 but that breaks other things 👀

kiaking:
Ah yeah, well too bad. This is hard one I guess 😅 Let's leave it for now 👍

kiaking (approved):

kiaking (approved):

PR #324, opened 2023-07-26
feat(input): enable input `font-family` custom property

### LINKED ISSUE #322: [Input] Enable font family customization
Let's add a new input related CSS variable (`--input-value-font-family`) that can adjust font family.

This is useful when we want to have mono spaced font family on textarea when we expect user to input for example JSON value.

cuebit:
resolves #322

### DIFF
--- a/lib/components/SInputYMD.vue
+++ b/lib/components/SInputYMD.vue
@@ -302,6 +302,7 @@ function createRequiredTouched(): boolean[] {
 
 .input {
   line-height: 24px;
+  font-family: var(--input-value-font-family, var(--font-family-number));
   font-feature-settings: "tnum";
   background-color: transparent;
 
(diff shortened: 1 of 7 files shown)


lib/components/SInputYMD.vue
```diff
@@ -302,6 +302,7 @@ function createRequiredTouched(): boolean[] {
 
 .input {
   line-height: 24px;
+  font-family: var(--input-value-font-family, var(--font-family-number));
```
cuebit, line 305: Added `--font-family-number` as a fallback. [`SInputHMS`](https://github.com/globalbrain/sefirot/blob/main/lib/components/SInputHMS.vue#L311) has it, is there any specific reason this didn't?

kiaking: Nope. I think because `SInputHMS` is older and I've forgot to add support to it 😅

kiaking (approved):
Thanks a bunch! I've added docs for this style 👍

PR #328, opened 2023-08-08
fix(desc): empty slot not showing empty state

kiaking:
I found that currently, if we pass null value to `<SDesc>` slots, it will not render `<SDEscEmpty>`. This PR fixes it by checking in the actual slot contents.

I've also added helper utilities to do this as well.

```html
<!-- This works. <SDescEmpty> renderd -->
<SDescText></SDescText>

<!-- This does not work -->
<SDescText>{{ undefined }}</SDescText>
```

### DIFF
--- a/docs/composables/utils.md
+++ b/docs/composables/utils.md
@@ -84,3 +84,39 @@ const menu = computedArray((arr) => {
   }
 })
 ```
+
+## `useHasSlotContent`
+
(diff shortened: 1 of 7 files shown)


brc-dd:
Looking at the tests, this will also treat `<SDescText><div></div></SDescText>` as empty, right?

brc-dd (approved):

kiaking:
@brc-dd Haven't tested but I think so...? At the moment, either way works in that case since we don't have that use case 😅

PR #331, opened 2023-08-16
fix(input-ymd): don't apply hover styles if input is disabled

### LINKED ISSUE #330: [SInputYmd] Hover styles are enabled even when `disabled` prop is true
The title is everything.
I found this bug while implementing a logic to disable form items in Deus.

Related #220

brc-dd:
closes #330

`SInputHMS` already have this:

https://github.com/globalbrain/sefirot/blob/97b1bf33b0003c2028192ab1ec899f17be41c694/lib/components/SInputHMS.vue#L279-L291


kiaking (approved):
Thanks a bunch!

PR #332, opened 2023-08-24
feat(desc): add `:dir` option

kiaking:
Adds `:dir` option to `SDesc`. This will align label and value horizontally.

To control the divider, I've moved divider to `SDescItem`. And added additional prop to disable it.

Finally, when setting `:dir` to `false`, specify `:labelWidth` to define the width of label area grid.

```html
<SDesc cols="2" gap="24" dir="row" label-width="96" :divider="false">
  ...
</SDesc>
```

---

Screenshot 2023-08-24 at 16 15 17

### DIFF
--- a/docs/components/desc.md
+++ b/docs/components/desc.md
@@ -87,6 +87,41 @@ At the top level, use `<SDesc>` and `<SDescItem>` to define the layout. `<SDesc>
 </SDesc>
 ```
 
+You may also define `:dir` to control the direction of the label and value stack. The default is set to `column` which stacks the label and value vertically. You may set it to `row` to stack them horizontally.
+
+When setting `:dir` to `row`, you should also specify `labelWidth` prop which will determine how much width should the label are have in the grid system. The value is in `px`.
+
+```ts
+interface Prop {
+  // Defaults to `column`.
+  dir?: 'column' | 'row'
+
+  // Defaults to 128.
+  labelWidth?: string | number
+}
+```
+
+```vue-html
+<SDesc cols="2" gap="24" dir="row" label-width="96">
+  ...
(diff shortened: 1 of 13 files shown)


brc-dd:
Is this expected for `cols = 1` (`dir = column`)?

https://github.com/globalbrain/sefirot/assets/40380293/9d99f742-8aa5-4a2d-aaed-4cf0b4aa94b6

IG it should be like this `cols = 1` (`dir = row`, `labelWidth < 0`):

image

kiaking:
> Is this expected for cols = 1 (dir = column)?

Ah shit, I'm gonna remove that option from story. That behaviour is expected. That's happening because "About" section has `span 2`. If all section has `span 1`, it works fine 👍

brc-dd (approved):

docs/components/desc.md
```diff
@@ -87,6 +87,41 @@ At the top level, use `<SDesc>` and `<SDescItem>` to define the layout. `<SDesc>
 </SDesc>
 ```
 
+You may also define `:dir` to control the direction of the label and value stack. The default is set to `column` which stacks the label and value vertically. You may set it to `row` to stack them horizontally.
+
+When setting `:dir` to `row`, you should also specify `labelWidth` prop which will determine how much width should the label are have in the grid system. The value is in `px`.
```
cuebit, line 92:
```suggestion
When setting `:dir` to `row`, you should also specify `labelWidth` prop which will determine how much width the label should have in the grid system. The value is in `px`.
```

cuebit (approved):
Couple of suggestions, otherwise LGTM 👍

kiaking:
Thanks a lot guys!

PR #336, opened 2023-08-28
feat(checkbox): allow `indeterminate` as the value

### LINKED ISSUE #335: [Checkbox] Support `indeterminate` state
While `SInputCheckbox` has binary state with `checked` prop, native `input[type="checkbox"]` supports 3rd state: `indeterminate`.

You will find what it mean by seeing [MDN document](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes).

With this state, we can implement bulk action UI in our systems (e.g. table UI with checkboxes and so on).

Before starting implementing, please summarize a draft spec and share it in this issue thread.

brc-dd:
closes #335

### DIFF
--- a/stories/components/SInputCheckbox.02_Indeterminate.story.vue
+++ b/stories/components/SInputCheckbox.02_Indeterminate.story.vue
@@ -1,0 +1,54 @@
+<script setup lang="ts">
+import SInputCheckbox from 'sefirot/components/SInputCheckbox.vue'
+import { computed, reactive } from 'vue'
+
+const title = 'Components / SInputCheckbox / 02. Indeterminate'
+const docs = '/components/input-checkbox'
+
+const options = reactive([
+  { label: 'Book', value: 1, modelValue: false },
+  { label: 'Diamonds (x2)', value: 2, modelValue: false },
+  { label: 'Obsidian (x4)', value: 3, modelValue: false }
+])
+
+const control = computed({
+  get() {
+    return options.every((option) => option.modelValue)
+      ? true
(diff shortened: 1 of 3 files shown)


kiaking:
I like the overall API design 👍

brc-dd:
There is a bit of issue. Technically `indeterminate` should work same as false. I'll update the logic a bit.

stories/components/SInputCheckbox.02_Indeterminate.story.vue
```diff
@@ -0,0 +1,54 @@
...
+  { label: 'Obsidian (x4)', value: 3, modelValue: false }
+])
+
+const control = computed({
+  get() {
+    return options.every((option) => option.modelValue)
+      ? true
+      : options.some((option) => option.modelValue)
+        ? 'indeterminate'
+        : false
```
brc-dd, line 20: This would also work the same:

```suggestion
    const count = options.filter((option) => option.modelValue).length
    return count === options.length
      ? true
      : count > 0 ? 'indeterminate' : false
```

brc-dd, line 20: This would work the same. Might even be faster. Is a bit less readable though 😅

```suggestion
    let count = 0
    options.forEach((option) => option.modelValue && count++)
    return count === options.length ? true : count > 0 ? 'indeterminate' : false
```

kiaking (approved):
Thanks a bunch! Working beautiful 👍

PR #338, opened 2023-09-11
chore: bump deps

brc-dd:
- bump deps
- drop ts and dev only stuff from peer deps
- bump pnpm -- peer deps and higest resolution workarounds are no longer needed
- switch testing,deploy to node 18 -- node 16 is reaching eol today
- fix typos
- fix types with histoire 0.17
- vitest c8 is renamed to v8
- provide default value for inject in `SDescItem` -- it was giving warning in tests
- change vitepress config to mts -- esm only package
- consistent type imports
- rest of the changes are done by eslint


kiaking (approved):
Didn't know Vuelidate was out of beta! Nice.

PR #339, opened 2023-09-11
fix(table): allow using virtual scroll with fixed column

### LINKED ISSUE #266: [Table] Fixed column is not working with virtual scroll
When using a fixed column in STable component like below, the column doesn't become sticky.
This is because some classes used in `vue-virtual-scroller` specify `overflow: hidden`.

So it might be better to have the options like `useVirtualScroll` on STable.

```vue
<script setup lang="ts">
  const table = useTable({
    orders: ['name', 'status', 'type', 'tags', 'createdAt'],
    ...
  })
</script>

<style scoped>
.table :deep(.col-name) {
  --table-col-position: sticky;
  --table-col-z-index: 50;
}
```
image

brc-dd:
closes #266

TODO:
- [x] revert story changes before merging

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -21,7 +21,7 @@ export interface Table<
   reset?: MaybeRef<boolean | undefined>
   borderless?: MaybeRef<boolean>
   loading?: MaybeRef<boolean | undefined>
-  virtualScroll?: MaybeRef<boolean | undefined>
+  rowSize?: MaybeRef<number | undefined>
   onPrev?(): void
   onNext?(): void
   onReset?(): void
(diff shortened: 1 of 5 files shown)


lib/composables/Table.ts
```diff
@@ -21,7 +21,7 @@ export interface Table<
   reset?: MaybeRef<boolean | undefined>
   borderless?: MaybeRef<boolean>
   loading?: MaybeRef<boolean | undefined>
-  virtualScroll?: MaybeRef<boolean | undefined>
+  rowSize?: MaybeRef<number | undefined>
```
brc-dd, line 24: this should include border width as well, so if the row height is 40px and there is 1px border, it should be 41px

kiaking (commented):
Nice!!! Hold on merging this one. We only have one place where we use this at the moment so we could test it there and see it works.

So I guess only change we need to make is to use option `virtualScroll` -> `rowSize` 👀

brc-dd:
rowSize is optional, you can simply remove virtualScroll

PR #341, opened 2023-09-12
fix(input): show info tooltip only on click

### LINKED ISSUE #259: [Input] Activate `:info` tooltip on click
- ref #255
- require https://github.com/globalbrain/sefirot/issues/256

Hover tooltip is [not that great](https://primer.style/design/guides/accessibility/tooltip-alternatives). Let's make this tooltip activate on click rather than hover.

Note that we shouldn't let it activate when clicking on the label because clicking label should usually focus the input. So, probably we should detach the tooltip from label text and make it live by it self.

brc-dd:
closes #259


kiaking (approved):
Looking great! Thanks 🙌

PR #344, opened 2023-09-13
feat(table): allow selecting rows

brc-dd:
API:

```ts
const data = shallowRef([...])
const selected = ref<string[]>([])

const table = useTable({
  // ...
  actions: computed(() => [
    {
      label: 'Delete selected',
      onClick: deleteSelected,
      labelMode: 'danger',
      show: selected.value.length > 0
    }
  ]),
  indexField: 'name'
})

function deleteSelected() {
  data.value = data.value.filter((i) => !selected.value.includes(i.name)) // dummy, delete from db or something
  selected.value = [] // clear selected after deletion
}

---

<STable :options="table" v-model:selected="selected" />
```

TODO:

- [x] reset story before merging


lib/components/STable.vue
```diff
@@ -20,6 +21,11 @@ import STableItem from './STableItem.vue'
 
 const props = defineProps<{
   options: Table
+  selected?: number[]
```
kiaking, line 24: Can we make this support `string[]` too? The unique identifier could be `string` in some cases 👀

stories/components/STable.01_Playground.story.vue
```diff
@@ -319,12 +319,19 @@ function updateTypeFilter(value: string) {
 function updateTagsFilter(value: string) {
   dropdownTagsSelected.value = xor(dropdownTagsSelected.value, [value])
 }
+
+const selected = ref([])
+
+watchEffect(() => {
```
kiaking, line 325: Memo: Don't forget to remove this before merging.

kiaking (commented):
Made few comments, but overall, I love it! very simple to use with good API 🎉

kiaking:
@brc-dd I've adjusted the style of the actions.

I wanted to make it look like this.

Screenshot 2023-09-19 at 15 34 30

So I also made "Reset filters" a button look as well. Then I realized now we don't need divider since button acts like it 👀 

So, I divided stats and actions array.

Also changed the text of total records. Removed `records` word when selected in not empty to reduce the overall text size.

Could you double check if I didn't break anything 🙏 

The functionality is working really smooth!

lib/components/STableHeaderActionItem.vue
```diff
@@ -0,0 +1,30 @@
...
+    <div class="STableHeaderActionItem">
+      <SButton
+        size="mini"
+        :mode="mode"
+        :label="label"
+        :label-mode="labelMode"
+        @click="onClick"
+      />
+    </div>
+  </template>
```
brc-dd, line 23:
```suggestion
  <div v-if="show" class="STableHeaderActionItem">
    <SButton
      size="mini"
      :mode="mode"
      :label="label"
      :label-mode="labelMode"
      @click="onClick"
    />
  </div>
```

stories/components/STable.01_Playground.story.vue
```diff
@@ -279,6 +281,17 @@ const table = useTable({
     }
   ]),
 
+  actions: computed(() => [
+    {
+      label: 'Delete selected',
+      onClick: deleteSelected,
+      mode: 'mute',
```
brc-dd, line 288:
```suggestion
```

(just testing)

brc-dd:
Working fine for me 🙌

brc-dd:
Should we deprecate the hasFilters + onReset? 👀 Like ignore that when actions is specified, so any new code should just pass that directly as action instead of passing through all those components?

kiaking:
> Should we deprecate the hasFilters + onReset? 👀 Like ignore that when actions is specified, so any new code should just pass that directly as action instead of passing through all those components? Might seem repetitive though, but would be more consistent.

Ah, Good point! You're right now we don't need it 👀 Let's mark it as deprecated 👍 Maybe adding comments to Table composable...?

kiaking:
We would have `isDirty` check anyway on the user land in order to actually reset the condition, so, I don't think it would add that much of extra work even we disabled `reset` option 👍

brc-dd:
note regarding the border fixes - https://cloudfour.com/thinks/the-math-behind-nesting-rounded-corners/

however, in our case due to 1px border width, it doesn't affect that much and one won't see the gap without zooming.

kiaking:
Ah, thanks!

PR #346, opened 2023-09-28
feat(input-file): allow specifying `accept` and `multiple` on the input element

brc-dd:


brc-dd:
We can revert the `multiple` related changes though. I don't exactly need it and it will break types in other projects.

kiaking (approved):
Nice! I think we can keep `multiple` option too. It might break the types but it would be easy fix 👍

PR #348, opened 2023-09-29
fix(table): sync scroll on filter/sort and remove hover listeners

brc-dd:


brc-dd:
Still not sure if it fixes all the issues. Try it locally once. I'm no longer able to reproduce the issues locally though.

kiaking (approved):
Thanks! yeah, let's try with this one and see if it gets fixed 👀

PR #351, opened 2023-10-03
fix(input): fix typo in class names

### LINKED ISSUE #350: [Input] The label font size is not correct
Seems like label font size is using the default value now but it should be `14px`. It's working fine on Histoire because it has font size 14px set globally for styling purpose but usually we have 16px as default on other projects, hence we should have explicit style for input label I guess.

Maybe we should add [Shared Input Styles](https://sefirot.globalbrains.com/styles/input-styles) for `--input-label-font-size: 14px` 👀 

Screenshot 2023-10-03 at 13 50 55

brc-dd:
closes #350


kiaking (approved):
Thanks!

PR #354, opened 2023-10-05
fix(input): limit help and error text width

### LINKED ISSUE #353: [Input] Limit the length of `help` and `error` text
Currently, it goes as long as the input when the text is long, and it would be hard to read.

I think we should have `max-width`. I think `640px` would do.

Screenshot 2023-10-05 at 19 30 53

brc-dd:
closes #353

Feel free to change it to 640px. I'd used 65ch because that's common with most of the other libraries like primer, tailwind/typography, most chrome and ms projects, etc.

### DIFF
--- a/lib/components/SInputBase.vue
+++ b/lib/components/SInputBase.vue
@@ -155,6 +155,7 @@ function getErrorMsg(validation: Validatable) {
 
 .help-error {
   width: 100%;
+  max-width: 65ch;
   margin: 0;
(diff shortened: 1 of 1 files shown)


kiaking:
Wow, never knew there was `ch` 😳  Confirmed this also works with Japanese. Though it's not really `65ch` but becomes `53ch`. Probably this is after all setting `width` I guess.

I say we should be using this method more. Better than `em` 👍

PR #360, opened 2023-10-23
feat(table): add action column type

brc-dd:
refer story for usage

still need to fix some things like auto-resize and sticky columns. probably can do them later. this should unblock nia

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -364,7 +364,7 @@ const options = useTable({
 Below are the complete list of options you may pass to the text type cell.
 
 ```ts
-export interface TableCellNumber {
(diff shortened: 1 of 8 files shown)


kiaking (approved):
Working great! I've:

- Adjusted the style just a bit (remove 2px gap).
- Made `mute` mode default for actions since they should most likely be mute all the time.
- Add docs for this cell type.

I've also lightly tested the position sticky but indeed it's working quite smooth, other than things you mentioned. Also need to figure out how to handle borders since if we make right column sticky, then border disappears 😅 

But I agree we can do that later. I'll create issues for those. Let's release this baby 🎉

PR #364, opened 2023-11-01
feat(action-list): add `<ActionList>` (#358)

### LINKED ISSUE #358: Add `<ActionList>`
- [Figma](https://www.figma.com/file/a6TKFm1rEZ3mty5XhooM9C/Sefirot?type=design&node-id=41%3A155&mode=design&t=Q2n1fSWvLVPfuO64-1)

`<ActionList>` is a component that list actionable list. Inspired by [GitHub Primer / Action list](https://primer.style/design/components/action-list).

ActionList_Simple_List

## Usage

```vue
<script setup lang="ts">
const options: ActionList = {
  type: 'buttons',
  items: [
    { leadIcon: IconActivity, text: 'Show activity', onClick: () => {} },
    { leadIcon: IconEye, text: 'Preview', onClick: () => {} },
    { leadIcon: IconTrash, text: 'Delete item', onClick: () => {} }
  ]
}
</script>

<template>
  <SActionList :options="options" />
</template>
```

## Props

```ts
interface Props {
  options: ActionList
}

interface ActionList {
  // In the future, we can add more types like `checkboxes`
  // or `radios` to have selectable menu.
  type?: 'buttons' // Default: 'buttons'

  items: MaybeRefOrGetter<ActionListItem[]>
}

interface ActionListItem {
  leadIcon?: IconifyIcon
  text: string

  // If the link is provided, use `<SLink>` to render this item.
  link?: string

  onClick(): void
}
```

## Future Plan

It would be nice to use this component within `<Dropdown>` to have unified UI like GitHub Primer does. Therefore, I'm adding `:type` here.

#### Within card block

ActionList_Simple_List_In_Card

#### Hover state

ActionList_Simple_List_In_Card_Hover

#### Without lead icon

ActionList_Simple_List_In_Card_No_Icon

kiaking:
- close #358 

Add `<ActionList>`. I have made the type simpler than I described on the issue. Well, this is all we need at the moment.

```ts
interface Props {
  list?: ActionList
}

type ActionList = ActionListItem[]

interface ActionListItem {
  leadIcon?: IconifyIcon
  text: string
  link?: string
  onClick?(): void
}
```

Also, I have made few css var adjustment to match the latest color system on Figma. `--c-mute` is deprecated in favor of new `--c-bg-mute-1` series. 

---

Screenshot 2023-11-01 at 16 06 48

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -48,6 +48,7 @@ function sidebar(): DefaultTheme.SidebarItem[] {
       text: 'Components',
       collapsed: false,
       items: [
+        { text: 'SActionList', link: '/components/action-list' },
         { text: 'SAvatar', link: '/components/avatar' },
         { text: 'SButton', link: '/components/button' },
         { text: 'SButtonGroup', link: '/components/button-group' },
--- a/docs/components/action-list.md
+++ b/docs/components/action-list.md
@@ -1,0 +1,106 @@
+<script setup lang="ts">
+import IconActivity from '@iconify-icons/ph/activity-bold'
+import IconEye from '@iconify-icons/ph/eye-bold'
+import IconTrash from '@iconify-icons/ph/trash-bold'
+import SActionList, { type ActionList } from 'sefirot/components/SActionList.vue'
+
+const list: ActionList = [
+  { leadIcon: IconActivity, text: 'Show activity' },
+  { leadIcon: IconEye, text: 'Preview' },
+  { leadIcon: IconTrash, text: 'Delete item' }
+]
+</script>
+
+# SActionList
+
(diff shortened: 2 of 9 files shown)


brc-dd (approved):

brc-dd:
Probably should use `markRaw(IconSomething)` -- Doesn't seem to be affecting in this case, but mostly Vue will log warning.

kiaking:
@brc-dd Ah, yeah. But that only happens when you set icon inside reactive right...? But that could happen as soon as we pass the object inside template...?

brc-dd:
Yeah. I'm not sure how reactivity exactly works with arrays passed as props. But if you make them computed or something, that most likely will throw warnings.

kiaking:
Make sense 👍

PR #365, opened 2023-11-01
feat(style): update bg color system

kiaking:
Update `--c-bg-elv-X` colors to match the latest Figma color system. It deprecates `--c-bg-soft`.


lib/styles/variables.css
brc-dd, line 1: maybe update the ``/* DEPRECATED: Use `--c-bg-soft`. */`` comments in this file?

brc-dd (approved):

lib/styles/variables.css
kiaking: Very good point! Updated 👍

PR #367, opened 2023-11-02
feat(styles): new color system

kiaking:
Huge PR that's introduces new color system. All older colors are been deprecated but still there, so it should not break existing apps.

- Moved deprecated styles to `variables-deprecated.css`
- Updated VitePress theme to match Sefirot theme as well 😎 

Screenshot 2023-11-02 at 18 39 08


brc-dd (approved):
Let's merge+release this and test it out on Marguerite 🚀

PR #369, opened 2023-11-06
fix(image): reset input element on remove

brc-dd:
input element wasn't being reset, so that change event is not triggered unless you're uploading different image

### DIFF
--- a/lib/components/SInputImage.vue
+++ b/lib/components/SInputImage.vue
@@ -71,6 +71,8 @@ function onFileSelect(e: Event) {
 }
 
 function onFileDelete() {
+  fileInput.value!.value = ''
+
   emit('update:model-value', null)
   emit('change', null)
 }


kiaking (approved):
Ahhhh make sense. thanks!

lib/components/SInputImage.vue
```diff
@@ -71,6 +71,8 @@ function onFileSelect(e: Event) {
 }
 
 function onFileDelete() {
+  fileInput.value!.value = ''
+
   emit('update:model-value', null)
   emit('change', null)
```
brc-dd, line 77: probably we don't need these 👀  changing value should trigger onFileSelect IG

brc-dd: ah, no, somehow that isn't happening. weird. keeping them should be fine IG.

kiaking: 👍

PR #371, opened 2023-11-07
fix(input-date): fix prop value to prevent warning

NozomuIkuta:
No related issues.

When `disabled` prop of `SInputDate` component is set true, Vue show the following warning:

> [Vue warn]: v-on with no argument expects an object value.

This PR fixes runtime type error and prevent the warning.

### DIFF
--- a/lib/components/SInputDate.vue
+++ b/lib/components/SInputDate.vue
@@ -86,7 +86,7 @@ function emitBlur() {
           autocomplete="off"
           :value="inputValue"
           :disabled="disabled"
-          v-on="disabled || inputEvents"
+          v-on="disabled ? {} : inputEvents"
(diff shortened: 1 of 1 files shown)


kiaking (commented):
Ah make sense why it is warning. Don't have much knowledge on the correct fix for `v-on`.

@brc-dd What do you think? If this looks good to you, I think we can merge 👍

brc-dd (approved):

brc-dd:
which date picker version are we using? on latest one, v-slot shouldn't be there, you simply need v-model 👀

brc-dd:
Yeah this should work fine 👍

kiaking:
Thanks!

PR #372, opened 2023-11-08
fix(table): don't grow actions column

brc-dd:
another heads up, like we make the last column `resizable: false` everywhere, while using actions, we should make the last "two" columns `resizable: false`.


kiaking (approved):
👍

PR #374, opened 2023-11-13
feat(table): allow conditionally showing action items

brc-dd:
This might be helpful in cases where we don't want to show delete button, say, in front of current user / only sysadmin.

Slightly breaking change - value is removed from onClick as it's always undefined.


kiaking (approved):
Make sense. Looking good to me 👍 Deel free to merge whenever you are ready.

I think we should have a feature that automatically sets `width` on action cell 🤔 It's a bit tricky because we have to set width for column too. I'll open another issue for this.

PR #381, opened 2023-11-13
feat(table): auto resize actions column

### LINKED ISSUE #377: [Table] Set actions column/cell width automatically
The width of table's actions cell can be determined by how many actions there are, so it would be cool if we can do that. A bit tricky since we need to set width for both column and cell.

The rule is this.

1. A button have width of `28px`.
2. There are no gap between buttons.
3. There should be `8px` padding left and right inside cell.

So, if we have 3 buttons, the width of the cell should be:

```ts
(button width 28px * 3) + (padding left 8px + padding right 8px) = 100px
```

When there are no buttons, set width to 40px. Which makes the cell square.

Screenshot 2023-11-13 at 16 24 12

brc-dd:
closes #377

---

Also adjusted the code to always left align the action cell items. It was looking weird when some of the icons are hidden:

Before:

image

After:

image

Not sure though which is more desirable.

---

Close  #382 if merging this. `show` is the recommended way if you want to hide some icons and use auto resize at the same time. This PR won't work when actions.cell is a function.

### DIFF
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -12,6 +12,7 @@ import {
   watch
 } from 'vue'
 import { type Table } from '../composables/Table'
+import { getTextWidth } from '../support/Text'
 import SInputCheckbox from './SInputCheckbox.vue'
 import SSpinner from './SSpinner.vue'
 import STableCell from './STableCell.vue'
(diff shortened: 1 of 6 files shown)


kiaking:
Nicely done!

> Also adjusted the code to always left align the action cell items. It was looking weird when some of the icons are hidden:

Yup, this is desirable I say. It should align left 👍 

> This PR won't work when actions.cell is a function.

Yeah we are so close. Confirmed that It does work with cell being function, unless user use the cell argument to do something... (passing in `value` and `record` arg here would be tough).

But I guess this limitation makes sense. It too complicated to tweak the cell type just for this.

One use case I have in mind is something [I commented on #832](https://github.com/globalbrain/sefirot/pull/382#issuecomment-1809348204), where we might wanna have different array rather than hiding action via `show` prop.

But I think for that case, we can do so by making `columns` reactive instead of making cell a function 👍 

```ts
const isAdmin = ref(false)

useTable({
  columns: computed(() => ({
    actions: {
      type: 'actions',
      actions: getActions()
    }
  }))
})

function getActions() {
  const actions = []
  
  actions.push({ icon: 'edit', onClick: () => {} })

  if (isAdmin.value) {
    actions.push({ icon 'delete', onClick: () => {} })
  }

  return actions
}

```

kiaking (approved):

brc-dd:
Yeah let's merge this and close #382 for now. If we need something complex, we probably can just specify the CSS variable.

kiaking:
Sounds good 👍

PR #382, opened 2023-11-13
Revert "feat(table): allow conditionally showing action items"

brc-dd:
Reverts globalbrain/sefirot#374

do something like instead:

```ts
    actions: {
      cell: (_, record) => { /* conditional logic here */ }
    }
```

### DIFF
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -510,7 +510,7 @@ interface TableCellActions {
   labelMode?: ColorModes
 
   // Callback function when the button is clicked.
-  onClick(value: any, record: any): void
+  onClick(record: any): void
 }
 
 type ColorModes =
--- a/docs/components/table.md
+++ b/docs/components/table.md
@@ -510,10 +510,7 @@ interface TableCellActions {
   labelMode?: ColorModes
 
   // Callback function when the button is clicked.
-  onClick(record: any): void
-
-  // Whether to show the button for the record.
-  show?(record: any): boolean
+  onClick(value: any, record: any): void
 }
 
 type ColorModes =
--- a/lib/components/STableCell.vue
+++ b/lib/components/STableCell.vue
@@ -102,6 +102,7 @@ const computedCell = computed<TableCell | undefined>(() =>
     />
     <STableCellActions
       v-else-if="computedCell.type === 'actions'"
+      :value="value"
       :record="record"
       :actions="computedCell.actions"
     />
--- a/lib/components/STableCell.vue
+++ b/lib/components/STableCell.vue
@@ -102,7 +102,6 @@ const computedCell = computed<TableCell | undefined>(() =>
     />
     <STableCellActions
       v-else-if="computedCell.type === 'actions'"
-      :value="value"
       :record="record"
       :actions="computedCell.actions"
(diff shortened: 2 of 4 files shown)


docs/components/table.md
```diff
@@ -510,10 +510,7 @@ interface TableCellActions {
   labelMode?: ColorModes
 
   // Callback function when the button is clicked.
-  onClick(record: any): void
-
-  // Whether to show the button for the record.
-  show?(record: any): boolean
+  onClick(value: any, record: any): void
```
brc-dd, line 513:
```suggestion
  onClick(record: any): void
```

lib/components/STableCell.vue
```diff
@@ -102,6 +102,7 @@ const computedCell = computed<TableCell | undefined>(() =>
     />
     <STableCellActions
       v-else-if="computedCell.type === 'actions'"
+      :value="value"
```
brc-dd, line 105:
```suggestion
```

lib/components/STableCellActions.vue
```diff
@@ -3,26 +3,26 @@ import { type TableCellAction } from '../composables/Table'
 import SButton from './SButton.vue'
 
 defineProps<{
+  value: any
```
brc-dd, line 6:
```suggestion
```

lib/components/STableCellActions.vue
```diff
@@ -3,26 +3,26 @@ import { type TableCellAction } from '../composables/Table'
...
+      v-for="(action, i) in actions"
+      :key="i"
+      size="mini"
+      type="text"
+      :mode="action.mode ?? 'mute'"
+      :icon="action.icon"
+      :icon-mode="action.iconMode"
+      :label="action.label"
+      :label-mode="action.labelMode"
+      @click="action.onClick(value, record)"
```
brc-dd, line 24:
```suggestion
      @click="action.onClick(record)"
```

lib/composables/Table.ts
```diff
@@ -189,8 +189,7 @@ export interface TableCellAction {
   iconMode?: Mode
   label?: string
   labelMode?: Mode
-  onClick(record: any): void
-  show?(record: any): boolean
+  onClick(value: any, record: any): void
```
brc-dd, line 192:
```suggestion
  onClick(record: any): void
```

brc-dd:
I'm not sure about this either. With the `cell.show` syntax I have an idea to fix #377 (-- we can calculate the width of labels and icons and add them + padding). Having cell functions would mess things a bit.

kiaking:
@brc-dd 
Yeah true. If we remove the actions from the array, it would be hard to calculate the desired width of the actions cell automatically 🤔 

It might be hard for everybody to follow the guideline on when to use `cell()` function and when to use `show` property, but I think `cell()` function best works when we know certain actions are not available all the time.

For example, if delete icon should only be visible to Admin, then guest users should only ever see 1 icon. In this case, we should just use `cell()` and add only 1 item in the array.

On the other hand, if the action is conditional based on each record, then we should be using `show` I guess.

So, I think having `show` at the moment make sense. What do you think 🤔

PR #384, opened 2023-11-14
feat(table): allow making right column sticky

### LINKED ISSUE #361: [Table] Add ability to make right column "sticky"
Currently, we can stick a column at left side of the table by doing:

```css
.table :deep(.col-name) {
  --table-col-position: sticky;
  --table-col-z-index: 50;
}
```

This only works for sticking the column on left side because:

1. There is no option to set position `right` on `<STableItem>`.
2. Left border disappears when scrolling since table cell only has border on right side.

We need to find a way to fix the above to enable right side sticky position.

brc-dd:
closes #361

### DIFF
--- a/stories/components/STable.01_Playground.story.vue
+++ b/stories/components/STable.01_Playground.story.vue
@@ -372,6 +372,13 @@ function updateTagsFilter(value: string) {
 .table :deep(.col-tags)      { --table-col-width: 192px; }
 .table :deep(.col-createdAt) { --table-col-width: 192px; }
 
+.table :deep(.col-actions) {
+  --table-col-position: sticky;
+  --table-col-z-index: 50;
+  --table-col-right: 0;
+  --table-col-border-left: 1px;
+}
+
 .table {
   margin-bottom: 16px;
 }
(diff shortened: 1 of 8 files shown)


brc-dd:
Looks like many other deprecated variables are also being used there 👀

kiaking:
Yeah we should replace it as we find any 👀

stories/components/STable.01_Playground.story.vue
```diff
@@ -372,6 +372,13 @@ function updateTagsFilter(value: string) {
 .table :deep(.col-tags)      { --table-col-width: 192px; }
 .table :deep(.col-createdAt) { --table-col-width: 192px; }
 
+.table :deep(.col-actions) {
+  --table-col-position: sticky;
+  --table-col-z-index: 50;
+  --table-col-right: 0;
```
kiaking, line 378: maybe it is safer to also add `--table-col-left: auto`?

brc-dd: ah yeah, wouldn't matter since last column will never reach left (at least till the screen size is not like 100px), but yeah let's add that.

brc-dd, line 378:
```suggestion
 --table-col-left: auto;
  --table-col-right: 0;
```

kiaking: Ahhh good point. Yes, this is position: sticky, not absolute. Got confused 😓  But yeah, why not be safe.

brc-dd: yeah updated it

kiaking (approved):
LGTM 👍

PR #387, opened 2023-11-14
chore: use release-it

brc-dd:


kiaking (approved):
Nice! Haven't been able to run it through but it looks good 👍

brc-dd:
We can also use https://github.com/semantic-release/semantic-release instead. semantic-release is more popular, but I haven't used it in a while 👀

kiaking:
Hmmm well I think this is fine though. I haven't use either so hard to tell the difference 😅

PR #395, opened 2023-11-20
chore: switch to vite 5

brc-dd:
Also,
- Fix SSR in `useImageSrcFromFile`
- Fix types in tests
- Enable search in docs

Histoire is kind of throwing some deprecation warnings and logging setup not found errors during build, but dev (tested hmr) and preview seem to work fine.


lib/composables/Image.ts
```diff
@@ -13,7 +13,13 @@ export interface ImageSrcFromFile {
 export function useImageSrcFromFile(
   file: MaybeRefOrGetter<File | string | null>
 ): ImageSrcFromFile {
-  const reader = new FileReader()
+  if (typeof document === 'undefined') {
```
kiaking, line 16: Should we add support function for this? 🤔  This could be common scenario when supporting SSR.

Something like:

```ts
// support/Utils.ts
// or `inBrowser`...? I think I saw other frameworks doing the
// similar thing so would like to follow though.
export function isBrowser(): boolean
```

brc-dd: We can just use isClient from vueuse 😅

kiaking: Oh OK hooray to VueUse 👍 They have everything 😆

kiaking (approved):

PR #396, opened 2023-11-20
feat: add and fix validation rules

NozomuIkuta:
No related issues.

- Fix `decimal` rule to deny hyphen-only input
  - For edge cases, add `decimalOrHyphen` rule to accept hyphen-only input, which is ex-`decimal` rule
  - Chaging `decimal` is a breaking change, but we will release it as "fix", because the old behavior is not what we expect
- Add new 4 rules: `positiveInteger`, `zeroOrPositiveInteger`, `negativeInteger`, and `zeroOrNegativeInteger`
  - We could define `zeroOr*Integer` validator and let `*Integer` rules accept zero, but I stick to define 2 primitive validators (`isZero` and `is*Integer`) for simplicity and maintainability, and rules of all 4 possible patterns to cover use cases that we might need in future


lib/validation/validators/isHyphen.ts
kiaking, line 1: I think we should remove `is` prefix. We don't use those for example with `decimal`, `checked`, etc.

I think these validators should have same name with rules.

NozomuIkuta: Fixed: e2f0812

I also added unit tests for newly added/modified rules (I didn't implement unit test for `positiveInteger` rule and `negativeInteger` rule because the results are same as their validators).

lib/validation/validators/negativeInteger.ts
```diff
@@ -0,0 +1,12 @@
+const basicRegExp = /^(?:\-(?:[1-9]+[0-9]*))$/
+const edgeCaseRegExp = /^(?:\-(?:[1-9]+[0-9]*\.0?))$/
+
+export function negativeInteger(value: string | number): boolean {
+  const _value = typeof value === 'number'
+    ? value.toString()
+    : edgeCaseRegExp.test(value)
+      ? Number(value).toString() // convert 1.0 to 1
+      : value
```
NozomuIkuta, line 9: It seems that `1.0` and `1.` are represented as `1` as string, so I couldn't come up with a way to distinguish integer `1` and decimal `1`.

Do you know any way to do that?

lib/validation/validators/positiveInteger.ts
```diff
@@ -0,0 +1,12 @@
+const basicRegExp = /^(?:\+?(?:[1-9]+[0-9]*))$/
+const edgeCaseRegExp = /^(?:\+?(?:[1-9]+[0-9]*\.0?))$/
+
+export function positiveInteger(value: string | number): boolean {
+  const _value = typeof value === 'number'
+    ? value.toString()
+    : edgeCaseRegExp.test(value)
+      ? Number(value).toString() // convert 1.0 to 1
+      : value
```
NozomuIkuta, line 9: Same as https://github.com/globalbrain/sefirot/pull/396/files#r1398978272

lib/validation/validators/negativeInteger.ts
kiaking: > It seems that 1.0 and 1. are represented as 1 as string

You want 1.0 to be treated as 1 right...? So can't we just do this?

```ts
const _value = Number(value).toString()
```

kiaking: Or I think this should work too...?

```ts
return Number(value) % 1 === 0
```

NozomuIkuta: Sorry, my description and current implementation are not so right. 🙏 

If I'm correct, `1.0` and `"1.0"` is decimal and the validator should not accept them , doesn't it?

We can detect `"1.0"` string to check the last part (i.e. check if the string ends with `"."` or `".0"`), but I can't find a way to distinguish `1` and `1.0` as number, because both `(1).toString()` and `(1.0).toString()` returns `"1"`.

If I'm correct, there is no way to distinguish `1` and `1.0` as number in JavaScript.

@brc-dd might know a way, though.

NozomuIkuta: Or, should we accept `1.0` and `"1.0"` as integer?

NozomuIkuta: > Or I think this should work too...?
> 
> return Number(value) % 1 === 0

Both `Number("1") % 1 === 0` and `Number("1.0") % 1 === 0` return `true` 😢

kiaking: Oh, I mean you have test that accept `1.0` 👀 

```ts
expect(positiveInteger('1.0')).toBe(true)
```

It really depends on this validators use case. I mean do we have to support string in the first place?

For example, if we use this validator with `SInputNumber` and user inputs `1.0`, I think it is OK (we store it as `1`). Same as things like `-0` (store it as `0`).

Maybe the string support is there to support Money type...(decimal)?

NozomuIkuta: > Oh, I mean you have test that accept 1.0 👀
> 
> ```ts
> expect(positiveInteger('1.0')).toBe(true)
> ```

Actually, I wanted to let the validator deny the value because I believe it's decimal.

> It really depends on this validators use case. I mean do we have to support string in the first place?

Well, you are right. In our use cases, only numbers will be passed to the validators.
And, if we come to need string support, we could add support for string without breaking changes.

I removed string type support (1199d6c), so that we can merge this PR and consider complex cases later.

kiaking (approved):

brc-dd (approved):

PR #402, opened 2023-11-24
feat(login): add `<SLoginPage>` component

### LINKED ISSUE #380: [LoginPage] Add `<LoginPage>` component
We use exact same login page for almost all apps. Would be nice to have this component built into Sefirot.

Should look something like this.

```html
<SLoginPage
  cover="https://cover-image.jpg/"
  cover-title="Tokyo in the Night"
  cover-photographer="John Doe"
  :actions="[
    { type: "google", onClick: () => { /* Handle login */ } }
  ]"
/>
```

NozomuIkuta:
Resolves #380

I implemented this component based on Barthromei's code base.


lib/components/SLoginPage.vue
```diff
@@ -0,0 +1,207 @@
...
+}
+
+export interface CoverPhotographer {
+  text: string
+  link: string
+}
+
+export interface Action {
+  type: 'google'
+  label?: string
```
NozomuIkuta, line 20: We might not need `label` prop to override login button text and just calculate it in `getActionLabel` function.
What do you think?

brc-dd:
Looking good to me. I'll test this once out with maru locally.

lib/components/SLoginPage.vue
kiaking: Yeah I agree, we don't need to pass in the text I think. I'll update it along with other style fixes 👍

kiaking (approved):
Updated few things.

- Use Brand logo from `@iconify-icons/ri` (remix icons).
  - This set contains good amount of branding icon so I think we can start using this instead of having ours.
- Updated style vars a bit.
- Removed `label` from `Action`.
- Updated docs to fit the latest styles (from `<SM>`).
- Use external link for image in Story so that we don't have to have our own.

@NozomuIkuta 
Looking good from my side! Feel free to merge when ever you are ready 👍

PR #404, opened 2023-11-29
feat(url): add `useUrlQuerySync` function

### LINKED ISSUE #397: [Url] Add `useUrlQuerySync` function
Often, we need to have a way to sync URL query params with ref or reactive. Would be nice to have good function does that.

## Example usage

Let's say we have this kind of api call.

```ts
// Option to pass in to the API request.
// Each value is the default value.
const options = reactive({
  page: 1,
  conditions: {
    name: 'John'
  }
})

const { data } = useQuery(options)

function showNextPage() {
  options.page++
}
```

And then, we would like to "sync" this option to URL query params so that we can create a direct link to this conditions.

```ts
const options = reactive({
  page: 1,
  conditions: {
    name: 'John'
  }
})

// Sync with query.
useUrlQuerySync(options, {
  casts: {
    page: Number
  }
})

const { data } = useQuery(options)

function showNextPage() {
  options.page++
}
```

## Detailed examples

Sync should be by-directional.

```ts
// when URL is
// `https://example.com`

const options = reactive({ page: 1 })

useUrlQuerySync(options)

// -> https://example.com
// If the query is default value, do not show params on URL.

options.page++

// -> https://example.com?page=2
// If query differs from initial value, add params.
```

```ts
// If user access page with
// https://example.com?page=5

const options = reactive({ page: 1 })

useUrlQuerySync(options)

options.page // <- 5. Should be synced with URL.
```

## Options

It should be able to cast or mutation the query params, because they are all `string`.

```ts
useUrlQuerySync(options, {
  casts: {
    page: (value) => Number(value)
  }
})
```

It should be able to exclude certain params from showing up on URL.

> [!NOTE]
> This option might be too much... we could always remove the values before passing them into the function 🤔 

```ts
useUrlQuerySync(options, {
  // Do not display `perPage` option in URL.
  exclude: ['perPage']
})
```

## Current implementation

We use something like this in our apps. Would be great to organize this one.

> [!WARNING]
> Note that argument is different from the above examples because I changed it.

```ts
export interface UseUrlQuerySyncOptions {
  state: Record<string, any>
  casts?: Record<string, (value: any) => any>
  exclude?: string[]
}

export function useUrlQuerySync({ state, casts = {}, exclude = [] }: UseQuerySyncOptions): void {
  const router = useRouter()
  const route = useRoute()

  const initialState = cloneDeep(state)

  syncQueryToState()

  watch(() => state, syncStateToQuery, { deep: true })

  function syncStateToQuery(): void {
    if (isEqual(state, initialState)) {
      router.replace({ query: {} })

      return
    }

    const newState = cloneDeep(state)

    exclude.forEach((path) => unset(newState, path))

    router.replace({ query: newState })
  }

  function syncQueryToState(): void {
    const newState = cloneDeep(state)

    mergeWith(newState, route.query, (stateValue, _queryValue, key) => {
      if (!isNaN(Number(key))) {
        return
      }

      if (stateValue === undefined) {
        return '__REMOVE__'
      }
    })

    mergeWith({}, newState, (_dummyValue, stateValue, key, _dummyParent, stateParent) => {
      if (stateValue === '__REMOVE__') {
        delete stateParent[key]
      }
    })

    for (const path in casts) {
      update(newState, path, casts[path])
    }

    exclude.forEach((path) => unset(newState, path))

    mergeWith(state, newState)
  }
}
```

## Alternatives

The alternative API ideas are welcome as well!

## Consideration

Maybe we could use [useUrlSearchParams](https://vueuse.org/core/useUrlSearchParams/#useurlsearchparams) from vueuse? But not sure if this is OK when using Vue Router 🤔 (directly modifying URL no going through router API).

brc-dd:
closes #397

### DIFF
--- a/lib/composables/Url.ts
+++ b/lib/composables/Url.ts
@@ -1,0 +1,98 @@
+import isPlainObject from 'lodash-es/isPlainObject'
+import { watch } from 'vue'
+import { useRoute, useRouter } from 'vue-router'
+
+export interface UseUrlQuerySyncOptions {
+  state: Record<string, any>
+  casts?: Record<string, (value: any) => any>
+  exclude?: string[]
+}
+
+export function useUrlQuerySync({
+  state,
+  casts = {},
+  exclude
+}: UseUrlQuerySyncOptions) {
+  const router = useRouter()
+  const route = useRoute()
+
+  const flattenInitialState = flattenObject(JSON.parse(JSON.stringify(state)))
+
+  setStateFromQuery()
+
+  watch(() => state, setQueryFromState, { deep: true })
+
+  function setStateFromQuery() {
+    const flattenState = flattenObject(state)
+    const flattenQuery = flattenObject(route.query)
+
+    Object.keys(flattenQuery).forEach((key) => {
+      if (exclude?.includes(key)) {
+        return
+      }
+
+      const value = flattenQuery[key]
+      if (value === undefined) {
+        return
+      }
+
+      const cast = casts[key]
+      flattenState[key] = cast ? cast(value) : value
+    })
+
+    Object.assign(state, unflattenObject(flattenState))
+  }
+
+  function setQueryFromState() {
+    const flattenState = flattenObject(state)
+    const flattenQuery = flattenObject(route.query)
+
+    Object.keys(flattenState).forEach((key) => {
+      if (exclude?.includes(key)) {
+        return
+      }
+
+      const value = flattenState[key]
+      const initialValue = flattenInitialState[key]
+
+      if (value === initialValue) {
+        delete flattenQuery[key]
+      } else {
+        flattenQuery[key] = value
+      }
+
+      if (flattenQuery[key] === undefined) {
+        delete flattenQuery[key]
+      }
+
+      router.replace({ query: unflattenObject(flattenQuery) })
+    })
+  }
+}
+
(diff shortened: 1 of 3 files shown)


lib/composables/Url.ts
```diff
@@ -0,0 +1,98 @@
+import isPlainObject from 'lodash-es/isPlainObject'
+import { watch } from 'vue'
+import { useRoute, useRouter } from 'vue-router'
+
+export interface UseUrlQuerySyncOptions {
+  state: Record<string, any>
+  casts?: Record<string, (value: any) => any>
```
kiaking, line 7: Maybe the value could string? Since this value only comes from url query 👀 (doesn't really matter I guess but...

```suggestion
  casts?: Record<string, (value: string) => any>
```

lib/composables/Url.ts
```diff
@@ -0,0 +1,98 @@
...
+  exclude?: string[]
+}
+
+export function useUrlQuerySync({
```
kiaking, line 11: Can we make the `state` to be an 1st arg? Since it's only required arg and if we don't need option, we can do:

```ts
useUrlQuerySync(options, { ... })
```

kiaking (approved):
@brc-dd 
I've made options optional and added the docs.

Could you review it and merge if it looks good 👍

PR #411, opened 2023-12-09
feat(power): add usePower composable

brc-dd:


brc-dd:
Merging this for now. Feel free to adjust the docs when you're available.

PR #412, opened 2023-12-15
fix(url): preserve reactivity while assigning state from query

brc-dd:


lib/composables/Url.ts
```diff
@@ -94,3 +94,29 @@ function unflattenObject(obj: Record<string, any>) {
...
+    dest.length = src.length
+    src.forEach((item, index) => {
+      if (typeof item === 'object' && item !== null) {
+        deepAssign(dest[index], item)
+      } else {
+        dest[index] = toRaw(item)
+      }
+    })
+  } else {
+    throw new TypeError('[deepAssign] src must be an object or array')
```
brc-dd, line 120: Haven't implemented the other types yet. We can implement them as they come.

kiaking: Sounds good 👍

kiaking (approved):
LGTM. Please merge when you are ready 👍

PR #414, opened 2023-12-15
fix(url): change cast callback value type to any

kiaking:
The value could be for example array, if the stored object is array. So, I guess it makes sense to make this `any` 👀 

Or should we scope this to something like string or object or array?

```ts
useUrlQuerySync(options, {
  casts: {
    // It is indeed array but type error.
    // No `map`, since value is typed as `string`.
    'tags': (v) => v.map(Number),
  }
})
```

### DIFF
--- a/lib/composables/Url.ts
+++ b/lib/composables/Url.ts
@@ -4,7 +4,7 @@ import { watch } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
 
 export interface UseUrlQuerySyncOptions {
-  casts?: Record<string, (value: string) => any>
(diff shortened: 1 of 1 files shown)


brc-dd (approved):

brc-dd:
Ah yeah. I forgot about arrays and objects. The end value is always string. Like even if you have something like `?foo[0]=1` it will be `{ foo: ['1'] }`  or maybe in case of objects -> `{ foo: { bar: { baz: 'always string' } } }`. Not sure though if we can represent this with anything better than `any` 👀

kiaking:
Yeah... well it's url. It must `any` 🤪

PR #415, opened 2023-12-18
feat(local-nav): add `<SLocalNav>`

kiaking:
Adds local nav component.

```vue
<script setup lang="ts">
const navTitle = [
  { text: 'ABC Company', link: '#' },
  { text: 'Series A' }
]

const navMenu = [
  [
    { icon: IconActivity, text: 'Overview', link: '#', active: true },
    { icon: IconCurrencyCircleDollar, text: 'Payments', link: '#' }
  ],
  [
    { icon: IconGear, text: 'Settings', link: '#' }
  ]
]
</script>

<template>
  <SLocalNav :title="navTitle" :menu="navMenu" />
</template>
```

---

Screenshot 2023-12-18 at 20 38 02

### DIFF
--- a/lib/components/SLocalNav.vue
+++ b/lib/components/SLocalNav.vue
@@ -1,19 +1,27 @@
 <script setup lang="ts">
+import { computed } from 'vue'
 import SLocalNavMenu, { type MenuItem } from './SLocalNavMenu.vue'
 import SLocalNavTitle, { type Title } from './SLocalNavTitle.vue'
 
 export type { Title, MenuItem }
 
-defineProps<{
+const props = defineProps<{
   title: Title[]
   menu?: MenuItem[][]
 }>()
+
+const normalizedMenu = computed(() => {
+  return props.menu?.reduce<MenuItem[][]>((carry, group) => {
+    group.length && carry.push(group)
+    return carry
+  }, []) ?? null
+})
 </script>
 
 <template>
-  <div class="SLocalNav" :class="{ 'has-menu': menu }">
+  <div class="SLocalNav" :class="{ 'has-menu': normalizedMenu }">
     <SLocalNavTitle :title="title" />
-    <SLocalNavMenu v-if="menu" :menu="menu" />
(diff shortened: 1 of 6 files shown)


lib/components/SLocalNav.vue
```diff
@@ -0,0 +1,39 @@
...
+  flex-direction: column;
+  gap: 8px;
+  padding: 16px 24px;
+
+  @media (min-width: 768px) {
+    padding: 16px 32px;
+  }
+}
+
+.SLocalNav:has(.SLocalNavMenu) {
```
brc-dd, line 32: This is new. Won't work with anything released before last year 😅

kiaking: Ah, I thought this one was way older 😓 I'll change it.

> Though, managing active is left to the component using this. Not sure how complex that will be 👀

It will add active status using `router-link-active` too. So if we set correct link, it should be automatic.

The `active` option was for more like demonstration purpose on Histoire and for escape hatch 👍

kiaking: Fixed!

lib/components/SLocalNav.vue
```diff
@@ -0,0 +1,39 @@
...
+  menu?: MenuItem[][]
+}>()
+</script>
+
+<template>
+  <div class="SLocalNav" :class="{ 'has-menu': menu }">
+    <SLocalNavTitle :title="title" />
+    <SLocalNavMenu v-if="menu" :menu="menu" />
+  </div>
+</template>
```
NozomuIkuta, line 18: You might want to add `has-menu` class and render `SLocalNavMenu` based on length of `menu` instead of `menu` existence, because `menu` array can be an empty under a certain condition and render empty `SLocalNavMenu` with 0 padding bottom, which I believe you don't want.

kiaking: Good point. Fixed 👍

NozomuIkuta (approved):

brc-dd (approved):

PR #417, opened 2023-12-19
feat(table): add option to control display count of avatars (#416)

### LINKED ISSUE #416: [Table] Add feature to control how many avatars should be shown in `STableCellAvatars`
Currently, `TableCellAvatars` are fixed limit only 3 avatars to be shown. There use case where we want show more users like 10 of them.

Let's add option to specify how many avatars should be shown.

```ts
export interface TableCellAvatars<V = any, R = any> extends TableCellBase {
  type: 'avatars'
  avatars: TableCellAvatarsOption[] | ((value: V, record: R) => TableCellAvatarsOption[])
  color?: 'neutral' | 'soft' | 'mute'

  // @default 2
  avatarCount?: number
  nameCount?: number
}
```

When count is set to 2. It should show 2 avatars, and when there is more than 2 avatars, show placeholder avatar as 3rd avatar (this is new change). Same goes for higher numbers.

For names, when it is set to 2, show 2 names, and then suffix with `+1`.

kiaking:
- close #416

Now we can control how many avatars do we wanna show on table cell. We can also set count to `0` to hide avatars or names.

```ts
const table = useTable({
  columns: {
    authors: {
      label: 'Authors',
      cell: (authors) => ({
        type: 'avatars',
        avatars: authors,
        avatarCount: 5,
        nameCount: 3
      })
    }
  }
})
```

Screenshot 2023-12-19 at 16 35 43

### DIFF
--- a/stories/components/STable.01_Playground.story.vue
+++ b/stories/components/STable.01_Playground.story.vue
@@ -130,6 +130,10 @@ const data = shallowRef([
     status: 'Published',
     type: 'Photo',
     width: 1280,
+    authors: [
+      { image: 'https://i.pravatar.cc/64?img=1', name: 'Jane Doe' },
+      { image: 'https://i.pravatar.cc/64?img=2', name: 'Daniel Green' }
+    ],
     createdAt: day('2022-10-10'),
     tags: ['Info', 'News', 'Latest']
   },
@@ -139,6 +143,9 @@ const data = shallowRef([
     status: 'Draft',
     type: 'Icon',
     width: 1280,
+    authors: [
+      { image: 'https://i.pravatar.cc/64?img=3', name: 'John Black' }
+    ],
     createdAt: day('2022-10-09'),
     tags: ['Info']
   },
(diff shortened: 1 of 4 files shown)


stories/components/STable.01_Playground.story.vue
```diff
@@ -367,6 +402,7 @@ function updateTagsFilter(value: string) {
 <style scoped>
 .table :deep(.col-name)      { --table-col-width: 160px; }
 .table :deep(.col-status)    { --table-col-width: 144px; }
+.table :deep(.col-authors)   { --table-col-width: 192px; }
```
brc-dd, line 405:
```suggestion
.table :deep(.col-authors)   { --table-col-width: 240px; }
```

brc-dd (approved):

brc-dd:
LGTM. The empty box is looking weird though 👀

NozomuIkuta (approved):
LGTM.

By the way, I agree with @brc-dd's suggestion.

kiaking:
I think this would work. Looks a bit like it got trimmed... but 🤔  What do you think,

Screenshot 2023-12-19 at 17 32 29

brc-dd:
yeah looks better I guess

brc-dd:
We can play around with margins if you want it more compact. For example, this is what github's look like (-11px margin for 20px img):

image

kiaking:
Haha, yeah this is tough one. Also GitHub avatars keep squashing as numbers grow. Somewhat interesting way to handle the "+1" part I think 👀

I guess we can keep it as-is. I think it is more important to show the full avatars so that we can recognize who those are 😳

brc-dd:
sounds good.

PR #421, opened 2023-12-20
feat(lang): add i18n feature

### LINKED ISSUE #359: Add I18N feature
It would be super cool if we can have default Japanese messages for things like validation error messages.

So, how about this design.

## Installation and usage

First we use Pinia to store locale for Sefirot. We can also use `provide/inject` but since we are already using Pinia for Snackbars, I think it would be easier.

To be able to tree shake, let user install lang as a plugin.

```ts
// main.ts
import { createSefirot } from '@globalbrain/seifort/lib/Sefirot'
import { ja } from '@globalbrain/seifort/lib/locales/ja'

// Create a new sefirot object for plugin installation.
// In case we need to hook something here in the future.
//
// By default, English is always available.
const sefirot = createSefirot({
  langs: { ja },
})

app.use(sefirot)
```

In user's app, set Sefirot locale when setting app locale. Directly to Sefirot Lang Store.

```ts
import { useLocaleStore } from '@globalbrain/seifort/lib/stores/Locale'

const localeStore = useLocaleStore()

localeStore.set('ja')
```

## Customization

User could create their own locale file to pass in as an plugin options. Maybe we should make the locale file flat, and make it easier to merge 🤔 

```ts
// main.ts
import { createSefirot } from '@globalbrain/seifort/lib/Sefirot'
import { ja } from '@globalbrain/seifort/lib/locales/ja'

const customJa = {
  'v_required': 'Custom validation message'
}

const sefirot = createSefirot({
  langs: { ja: { ...ja, ...customJa } },
})

app.use(sefirot)
```

kiaking:
- close #359

Here we go. I've used `Trans` inspired by Laravel 👀 Now, we have everything in Sefirot, so I don't think we need to create Lang related things in App except for switching locale and stuff.

This PR only contains localization for `required` rule, but it shouldn't have breaking change, and localization support can be added incrementally.

```ts
// In App.vue
import { provideLang } from 'sefirot/composables/Lang'

const lang = getAppLang()

provideLang(lang)
```

```ts
// Validation rule definition.
import { createRule } from '../Rule'
import { required as baseRequired } from '../validators'

export const message = {
  en: 'The field is required.',
  ja: 'この項目は必須です。'
}

export function required(msg?: string) {
  return createRule({
    message: ({ lang }) => msg ?? message[lang],
    validation: (value) => baseRequired(value)
  })
}
```


lib/composables/Lang.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const SefirotLangKey = 'sefirot-lang-key'
+
+export function provideLang(lang: Lang) {
+  provide(SefirotLangKey, lang)
+}
+
+export function useLang(): Lang {
+  return inject(SefirotLangKey) ?? 'en'
```
brc-dd, line 21:
```suggestion
  return inject(SefirotLangKey, 'en')
```

it'll throw otherwise

lib/validation/rules/required.ts
```diff
@@ -1,8 +1,14 @@
-import { required as baseRequired, helpers } from '@vuelidate/validators'
+import { createRule } from '../Rule'
+import { required as baseRequired } from '../validators'
+
+export const message = {
```
brc-dd, line 4: this will cause issues, as we are re-exporting everything from /rules/index.ts (every rule will be exporting messages)

kiaking: Ah, you are right. Hmmmm... It might be useful to have messages exported so...

Should we change `index.ts` to selectively export only the function? Then when we want to access the message, we can directly import { message } from `../{rule}` 👀

brc-dd: We should actually remove barrel exports. It messes up vite (slow HMR + bad treeshaking).

kiaking: Yeah. I'll do that 👀

brc-dd:
```ts
export { and, not, or } from '@vuelidate/validators'
export { checked } from './checked'
export { decimal } from './decimal'
export { decimalOrHyphen } from './decimalOrHyphen'
export { email } from './email'
export { fileExtension } from './fileExtension'
export { hms } from './hms'
export { maxFileSize } from './maxFileSize'
export { maxLength } from './maxLength'
export { maxTotalFileSize } from './maxTotalFileSize'
export { maxValue } from './maxValue'
export { minLength } from './minLength'
export { minValue } from './minValue'
export { month } from './month'
export { negativeInteger } from './negativeInteger'
export { positiveInteger } from './positiveInteger'
export { required } from './required'
export { requiredHms } from './requiredHms'
export { requiredIf } from './requiredIf'
export { requiredYmd } from './requiredYmd'
export { rule } from './rule'
export { url } from './url'
export { ymd } from './ymd'
export { zeroOrNegativeInteger } from './zeroOrNegativeInteger'
export { zeroOrPositiveInteger } from './zeroOrPositiveInteger'
```

paste this 😅

kiaking: Done!

brc-dd (approved):

PR #430, opened 2024-01-05
feat(url): support ref as state

### LINKED ISSUE #429: [Url] Support `MaybeRef` on `useUrlQuerySync`
Currently, `useUrlQuerySync` only accept reactive object as a state, but would be cool if we can pass in `MaybeRefOrGetter`.

brc-dd:
closes #429

### DIFF
--- a/lib/composables/Url.ts
+++ b/lib/composables/Url.ts
@@ -1,6 +1,6 @@
 import isEqual from 'lodash-es/isEqual'
 import isPlainObject from 'lodash-es/isPlainObject'
-import { watch } from 'vue'
+import { type MaybeRef, isRef, toValue, watch } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
 
 export interface UseUrlQuerySyncOptions {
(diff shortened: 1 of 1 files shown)


lib/composables/Url.ts
```diff
@@ -39,11 +44,14 @@ export function useUrlQuerySync(
       flattenState[key] = cast ? cast(value) : value
     })
 
-    deepAssign(state, unflattenObject(flattenState))
+    deepAssign(
+      isRef<Record<string, any>>(state) ? state.value : state,
```
kiaking, line 48: Just curious, probably I'm missing something but can't we use `toValue` here?

brc-dd: Yeah I was thinking that too. I'm not sure if assignments work with it. I'll take a look.

brc-dd: I initially thought assignment might not work with that. Seems to be working fine. Updated it.

kiaking: Nice. Yeah I was curious too so went ahead and checked the code but this ternary is what exactly Vue is doing 👍 
https://github.com/vuejs/core/blob/main/packages/reactivity/src/ref.ts

```ts
export function toValue<T>(source: MaybeRefOrGetter<T> | ComputedRef<T>): T {
  return isFunction(source) ? source() : unref(source)
}

export function unref<T>(ref: MaybeRef<T> | ComputedRef<T>): T {
  return isRef(ref) ? ref.value : ref
}
```

brc-dd: Yeah. I was just looking at that too 😅 We don't need toValue either. unref should work fine here and avoid extra checks for function.

kiaking: Very good point.

kiaking (approved):
@brc-dd 
LGTM! Please feel free to merge when you are ready, once merged, could you make an another release? 🙏

brc-dd:
released

kiaking:
Thanks!

PR #432, opened 2024-01-05
feat(lang): add useBrowserLang function

### LINKED ISSUE #422: [Lang] add `useBrowserLang`
Get user locale from browser settings. This is GB specific feature on how we handle the param. What we do is this.

```ts
import { useNavigatorLanguage } from '@vueuse/core'

const { language: browserLang } = useNavigatorLanguage()

return (browserLang.value === 'ja' || browserLang.value === 'ja-JP') ? 'ja' : 'en'
```

```ts
// API
function useBrowserLang(): Ref<Lang>
```

brc-dd:
closes #422

I don't think we need reactivity here. `useNavigatorLanguage` changes the language ref when a browser fires `languagechange` event, which in our case doesn't happen IG (you need to change preferred languages in browser settings to trigger it.)

### DIFF
--- a/lib/composables/Lang.ts
+++ b/lib/composables/Lang.ts
@@ -33,7 +33,7 @@ export function useTrans<T>(messages: TransMessages<T>): Trans<T> {
   }
 }
 
-export function useBrowserLang() {
+export function useBrowserLang(): Lang {
   const lang = navigator.language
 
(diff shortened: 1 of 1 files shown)


kiaking (approved):
Didn't know there was such an simple API 😆  Thanks a lot!

PR #435, opened 2024-01-10
feat(http): allow overriding fetch handler

brc-dd:
Bit of a breaking change:

```ts
// earlier -- Http.base = '/foo' -- now:

Http.config({
  baseURL: '/foo'
})
```

### DIFF
--- a/lib/composables/Api.ts
+++ b/lib/composables/Api.ts
@@ -30,7 +30,7 @@ export function useQuery<Data = any>(
   const loading = ref(false)
   const data = ref<Data | undefined>()
 
-  if (options.immediate !== false) {
+  if (options.immediate !== false && !import.meta.env.SSR) {
     execute()
   }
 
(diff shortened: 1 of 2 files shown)


kiaking:
I think it would be cooler if we can hide ofetch beneath Http 🤔  How about we add config method?

```ts
Http.config({
  base: '...',
  xsrfUrl: false
})
```

brc-dd:
Ah, but need to replace ofetch entirely in nuxt. Maybe this: ?

```ts
Http.config({
  base: '...',
  xsrfUrl: false,
  $fetch: ... // when necessary?
})
```

cuebit:
> Ah, but need to replace ofetch entirely in nuxt. Maybe this: ?
> 
> 
> 
> ```ts
> 
> Http.config({
> 
>   base: '...',
> 
>   xsrfUrl: false,
> 
>   $fetch: ... // when necessary?
> 
> })
> 
> ```

Yea this makes more sense. Probably ideal to allow passing an implementation than conditional operation where dependencies are concerned.

brc-dd:
updated

kiaking (approved):
Nice! Looking smooth 👍

PR #439, opened 2024-01-15
feat: add localize Vietnamese for validation&components

tachibana-shin:


kiaking:
Thank you so much for the contribution!

However, we might be able to provide some hooks in future if there is significant demand, but storing all the translations in the core will significantly impact the bundle size. We have both English and Japanese because those are the 2 langs we almost always use 🙏

PR #440, opened 2024-01-15
refactor(validation): change validator argument types

### LINKED ISSUE #438: [Validation] Make `value` argument of validators functions of type `unknown`
This issue is a follow-up of https://github.com/globalbrain/sefirot/pull/437#discussion_r1449876683.

NozomuIkuta:
Resolves #438

This PR changes first argument of validator functions are changed as of type `unknown`.

I made validator functions return `false` if type of the passed argument is not expected by them, but should I make them return `true` instead?

Other details are written as inline comments.

### DIFF
--- a/lib/support/Utils.ts
+++ b/lib/support/Utils.ts
@@ -25,7 +25,7 @@ export function isFile(value: unknown): value is File {
 }
 
 export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): value is Ymd {
-  if (value === null || typeof value !== 'object') {
+  if (!isObject(value)) {
     return false
   }
 
@@ -34,11 +34,11 @@ export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): va
       keys.push(YmdMap[type])
       return keys
     }, [])
-    .every((v) => Object.hasOwn(value, v))
+    .every((key) => value[key] === null || isString(value[key]) || isNumber(value[key]))
 }
 
 export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): value is Hms {
-  if (value === null || typeof value !== 'object') {
+  if (!isObject(value)) {
     return false
   }
 
@@ -47,5 +47,5 @@ export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): va
       keys.push(HmsMap[type])
       return keys
     }, [])
-    .every((v) => Object.hasOwn(value, v))
+    .every((key) => value[key] === null || isString(value[key]) || isNumber(value[key]))
 }
--- a/lib/support/Utils.ts
+++ b/lib/support/Utils.ts
@@ -1,4 +1,4 @@
-import { type Hms, HmsMap, type Ymd, YmdMap } from './Day'
+import { type Hms, HmsMap, type HmsType, type Ymd, YmdMap, type YmdType } from './Day'
 
 export function isNullish(value: unknown): value is undefined | null {
   return value === null || value === undefined
@@ -24,18 +24,28 @@ export function isFile(value: unknown): value is File {
   return value instanceof File
 }
 
-export function isYmd(value: unknown): value is Ymd {
-  return (
-    value !== null
-    && typeof value === 'object'
-    && Object.values(YmdMap).every((v) => Object.hasOwn(value, v))
-  )
+export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): value is Ymd {
+  if (value === null || typeof value !== 'object') {
+    return false
+  }
+
+  return required
+    .reduce<string[]>((keys, type) => {
+      keys.push(YmdMap[type])
+      return keys
+    }, [])
+    .every((v) => Object.hasOwn(value, v))
 }
 
-export function isHms(value: unknown): value is Hms {
-  return (
-    value !== null
-    && typeof value === 'object'
-    && Object.values(HmsMap).every((v) => Object.hasOwn(value, v))
-  )
+export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): value is Hms {
+  if (value === null || typeof value !== 'object') {
+    return false
+  }
+
+  return required
+    .reduce<string[]>((keys, type) => {
+      keys.push(HmsMap[type])
+      return keys
+    }, [])
+    .every((v) => Object.hasOwn(value, v))
 }
--- a/lib/support/Utils.ts
+++ b/lib/support/Utils.ts
@@ -34,7 +34,7 @@ export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): va
       keys.push(YmdMap[type])
       return keys
     }, [])
-    .every((key) => value[key] === null || isString(value[key]) || isNumber(value[key]))
+    .every((key) => value[key] === null || isNumber(value[key]))
 }
 
 export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): value is Hms {
@@ -47,5 +47,5 @@ export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): va
       keys.push(HmsMap[type])
       return keys
     }, [])
-    .every((key) => value[key] === null || isString(value[key]) || isNumber(value[key]))
+    .every((key) => value[key] === null || isString(value[key]))
 }
--- a/lib/support/Utils.ts
+++ b/lib/support/Utils.ts
@@ -1,5 +1,3 @@
-import { type Hms, HmsMap, type HmsType, type Ymd, YmdMap, type YmdType } from './Day'
-
 export function isNullish(value: unknown): value is undefined | null {
   return value === null || value === undefined
 }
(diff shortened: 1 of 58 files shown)


tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "sourceMap": true,
     "skipLibCheck": true,
 
-    "lib": ["es2020", "dom", "dom.iterable"],
+    "lib": ["es2022", "dom", "dom.iterable"],
```
NozomuIkuta, line 15: This change is to use `Object.hasOwn()` static method.
If this change should be avoided, I will use `hasOwnProperty()` instance method instead.

lib/validation/validators/maxLength.ts
```diff
@@ -1,3 +1,9 @@
-export function maxLength(value: string, length: number): boolean {
+import { isArray, isString } from '../../support/Utils'
+
+export function maxLength(value: unknown, length: number): boolean {
+  if (!(isString(value) || isArray(value))) {
+    return false
+  }
+
   return value.length <= length
 }
```
NozomuIkuta, line 9: Should this validator accept any value that has `length` property besides string and array?

lib/validation/validators/minLength.ts
```diff
@@ -1,3 +1,9 @@
-export function minLength(value: string, length: number): boolean {
+import { isArray, isString } from '../../support/Utils'
+
+export function minLength(value: unknown, length: number): boolean {
+  if (!(isString(value) || isArray(value))) {
+    return false
+  }
+
   return value.length >= length
 }
```
NozomuIkuta, line 9: Should this validator accept any value that has length property besides string and array?

lib/validation/validators/negativeInteger.ts
```diff
@@ -1,3 +1,9 @@
-export function negativeInteger(value: number): boolean {
+import { isNumber } from '../../support/Utils'
+
+export function negativeInteger(value: unknown): boolean {
+  if (!isNumber(value)) {
+    return false
+  }
+
   return Number.isInteger(value) && value < 0
 }
```
NozomuIkuta, line 9: `isNumber` guard is needed because `Number.isInteger()` static method just returns boolean without type predicate (i.e. `value is number`).

tests/validation/validators/hms.spec.ts
```diff
@@ -8,10 +8,19 @@ describe('validation/validators/hms', () => {
...
   })
 
   it('should validate only the given types', () => {
     expect(hms({ hour: '23', minute: '59', second: '60' }, ['h', 'm'])).toBe(true)
     expect(hms({ hour: '23', minute: '60', second: '60' }, ['h', 'm'])).toBe(false)
+    expect(hms({ hour: '23', minute: '59' }, ['h', 'm'])).toBe(true)
+    expect(hms({ hour: '23', minute: null }, ['h', 'm'])).toBe(true)
+
+    expect(hms({ hour: '23' }, ['h', 'm'])).toBe(false)
+    expect(hms({ hour: '23', minute: undefined }, ['h', 'm'])).toBe(false)
```
NozomuIkuta, line 24: I believe these test cases are expected and don't break existing behavior.

tests/validation/validators/requiredHms.spec.ts
```diff
@@ -5,11 +5,18 @@ describe('validation/validators/requiredHms', () => {
...
 
   test('validates only given types', () => {
-    const data = { hour: '01', minute: '02', second: null }
+    expect(requiredHms({ hour: '01', minute: '02', second: null }, ['h', 'm'])).toBe(true)
+    expect(requiredHms({ hour: '01', minute: '02' }, ['h', 'm'])).toBe(true)
 
-    expect(requiredHms(data, ['h', 'm'])).toBe(true)
+    expect(requiredHms({ hour: '01' }, ['h', 'm'])).toBe(false)
+    expect(requiredHms({ hour: '01', minute: undefined }, ['h', 'm'])).toBe(false)
+    expect(requiredHms({ hour: '01', minute: null }, ['h', 'm'])).toBe(false)
```
NozomuIkuta, line 20: I believe these test cases are expected and don't break existing behavior.

tests/validation/validators/requiredYmd.spec.ts
```diff
@@ -5,11 +5,18 @@ describe('validation/validators/requiredYmd', () => {
...
 
   test('validates only given types', () => {
-    const data = { year: 2000, month: 12, date: null }
+    expect(requiredYmd({ year: 2000, month: 12, date: null }, ['y', 'm'])).toBe(true)
+    expect(requiredYmd({ year: 2000, month: 12 }, ['y', 'm'])).toBe(true)
 
-    expect(requiredYmd(data, ['y', 'm'])).toBe(true)
+    expect(requiredYmd({ year: 2000 }, ['y', 'm'])).toBe(false)
+    expect(requiredYmd({ year: 2000, month: undefined }, ['y', 'm'])).toBe(false)
+    expect(requiredYmd({ year: 2000, month: null }, ['y', 'm'])).toBe(false)
```
NozomuIkuta, line 20: I believe these test cases are expected and don't break existing behavior.

tsconfig.json
brc-dd: I don't think hasOwn/hasOwnProperty is even needed there. That logic should probably be

```ts
.every((v) => value[v] === null || typeof value[v] === 'number') // === 'string' in isHms
```

lib/validation/validators/maxLength.ts
brc-dd: Wouldn't matter much in my opinion. We aren't using this for anything else than string.

tsconfig.json
NozomuIkuta: At the point of `value[v]`, type of value is `object` and TypeScript throws an type error:

> Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
> No index signature with a parameter of type 'string' was found on type '{}'.

This is why I thought it is needed to use either of `hasOwn/hasOwnProperty`.
The reason I used `hasOwn` over `hasOwnProperty` is just it's free from prototype pollution.

Is there any way to resolve this type error, or do you think it's reasonable enough to use `as any`?

brc-dd: Casting value as any is fine IG.

brc-dd: You can make types of reduce stricter too:

```ts
export function hasProp<T extends string>(value: Record<string, unknown>, prop: T): value is { [key in T]: unknown } {
  return Object.hasOwn(value, prop)
}

export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): value is Ymd {
  if (!isObject(value)) {
    return false
  }

  return required
    .reduce<(keyof Ymd)[]>((keys, type) => {
      keys.push(YmdMap[type])
      return keys
    }, [])
    .every((v) => hasProp(value, v) && isNumber(value[v]))
}

export function isHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): value is Hms {
  if (!isObject(value)) {
    return false
  }

  return required
    .reduce<(keyof Hms)[]>((keys, type) => {
      keys.push(HmsMap[type])
      return keys
    }, [])
    .every((v) => hasProp(value, v) && isString(value[v]))
}
```

NozomuIkuta: > BTW, if you use isObject it won't throw any TS errors:

Oh yes, it has type predicate.
Thank you, fixed: 4a4eddb

lib/support/Utils.ts
```diff
@@ -21,3 +23,29 @@ export function isObject(value: unknown): value is Record<string, any> {
...
+  if (!isObject(value)) {
+    return false
+  }
+
+  return required
+    .reduce<string[]>((keys, type) => {
+      keys.push(YmdMap[type])
+      return keys
+    }, [])
+    .every((key) => value[key] === null || isString(value[key]) || isNumber(value[key]))
```
brc-dd, line 37: Why allow strings here? The type only has `number | null` 👀

NozomuIkuta: Ah, it's just a copy-paste issue.
`Ymd` and `Hms` has different types: `number | null` and `string | null`.
Fixed: 2a1bbd1

lib/support/Utils.ts
```diff
@@ -21,3 +23,29 @@ export function isObject(value: unknown): value is Record<string, any> {
 export function isFile(value: unknown): value is File {
   return value instanceof File
 }
+
+export function isYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): value is Ymd {
```
kiaking, line 27: Let's move this to `support/Day`

lib/validation/validators/minLength.ts
kiaking: No let's stick with only string and array 👍

lib/support/Utils.ts
NozomuIkuta: Moved: bd8778c

lib/validation/validators/maxLength.ts
NozomuIkuta: @kiaking Any thoughts?

NozomuIkuta: Ah, sorry you already replied to another thread 🙌

kiaking (approved):

PR #446, opened 2024-01-22
feat(day): allow passing dayjs object to createYmd and createHms

### LINKED ISSUE #445: [Day] Add helper method to generate `Ymd` and `Hms` from `day` object
Would be nice to have a function that can generate `Ymd` and `Hms` from the `day` object 😳 

```ts
const d = day() // Say today is 1985-10-10

createYmdFromDay(d)

/*
 * Note that we must add 1 to the month because
 * `day.month()` starts from 0 but Ymd needs 1.
 * {
 *   year: 1985,
 *   month: 10,
 *   date: 10
 * }
 */ 
```

brc-dd:
closes #445

### DIFF
--- a/lib/support/Day.ts
+++ b/lib/support/Day.ts
@@ -60,7 +60,7 @@ export function tz(input?: Input, timezone?: string): Day {
 /**
  * Creates a new `Ymd` object.
  */
-export function createYmd(year: number, month: number, date: number): Ymd
+export function createYmd(year?: number | null, month?: number | null, date?: number | null): Ymd
 export function createYmd(day: Day): Ymd
 export function createYmd(
   yearOrDay: number | Day | null = null,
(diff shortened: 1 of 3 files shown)


lib/support/Day.ts
```diff
@@ -60,30 +60,50 @@ export function tz(input?: Input, timezone?: string): Day {
...
+      date
+    }
+  }
+
   return {
-    year,
-    month,
-    date
+    year: yearOrDay.year(),
+    month: yearOrDay.month(),
```
kiaking, line 80: I think we need to add `1` here since `day.month()` returns `0` for January. I might be wrong 👀

brc-dd: updated

kiaking (approved):

PR #449, opened 2024-01-24
chore: fix pnpm action

NozomuIkuta:
Complement #448
I just added pnpm version and add `run_install: false` because the following action do that.


kiaking (approved):

brc-dd:
Ah this is not needed actually. The action picks up from packageManager field and run_install is false by default.

PR #450, opened 2024-01-24
Revert "chore: fix pnpm action"

brc-dd:
Reverts globalbrain/sefirot#449 (refer https://github.com/globalbrain/sefirot/pull/449#issuecomment-1907647535)

Let's avoid having to change multiple places while updating deps.

Docs - https://github.com/pnpm/action-setup?tab=readme-ov-file#inputs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,6 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v2
-        with:
-          version: 8.14.3
-          run_install: false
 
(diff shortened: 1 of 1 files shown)


kiaking (approved):

NozomuIkuta:
Oh, thank you for pointing it out.

I added the lines following Deus.
So, we will be able to remove it from Deus and other products' codebase too, right?

brc-dd:
> So, we will be able to remove it from Deus and other products' codebase too, right?

Yeah

PR #451, opened 2024-01-24
chore: bump deps, simplify ci, fix docs, suppress inject warnings in tests

brc-dd:


kiaking (approved):

brc-dd:
Wait, there are bunch of inject can only be used in component warnings because of the recent changes in validation. Fixing those too.

PR #453, opened 2024-01-25
refactor(input-radios): improve types

brc-dd:
Nick gave me some idea in vue internal discord. 

Still need some way to avoid `null as NullValue`. Will see. After that we can implement generics for dropdown, etc.


kiaking (approved):
Nice 🙌

PR #460, opened 2024-01-31
fix(desc): show non-image avatar correctly

NozomuIkuta:
Fix #458

I found `SDescAvatar` shows avatar only when it's truthy, which is different from how `SAvatar` works.
So I removed if-condition for `SAvatar` to have a control on what to show: image or initial of name.

@kiaking 

If this is intended, please let me know and close this PR.


kiaking:
Ahhh good catch! thansk!

kiaking (approved):

PR #465, opened 2024-02-05
fix(api): execute query inside onMounted to avoid SSR hydration issues

brc-dd:


lib/composables/Utils.ts
```diff
@@ -61,3 +62,11 @@ export function useSlotValue(name = 'default'): ComputedRef<string | null> {
     return v !== '' ? v : null
   })
 }
+
+export function tryOnMounted(fn: () => void): void {
```
brc-dd, line 66: Not using vueuse's version because it executes the function in SSR which is not expected. Also this versions aligns better with vue as vue too doesn't run onMounted's callback in SSR.

kiaking (approved):
LGTM 👍  Please merge at your convenience.

PR #466, opened 2024-02-06
feat(api): allow watching options in useQuery

brc-dd:
Before:

```ts
export function usePostPage(options: MaybeRefOrGetter<UsePostPageOptions>) {
  const query = useQuery(async (http) => {...})

  watch(() => toValue(options), query.execute, { deep: true })

  return query
}
```

After:

```ts
export function usePostPage(options: MaybeRefOrGetter<UsePostPageOptions>) {
  return useQuery(async (http) => {...}, { watch: () => toValue(options) })
}
```

Nuxt's `useAsyncData` also supports passing watch sources, so `useQuerySsr` already supports this.

### DIFF
--- a/lib/composables/Api.ts
+++ b/lib/composables/Api.ts
@@ -1,4 +1,4 @@
-import { type Ref, ref } from 'vue'
+import { type Ref, type WatchSource, ref, watch } from 'vue'
 import { Http } from '../http/Http'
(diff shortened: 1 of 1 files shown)


kiaking (approved):
Ahhh smart 👍  I like it.

PR #467, opened 2024-02-12
fix: many bugs with useUrlQuerySync

brc-dd:
rewrote it 🫠

The caveats are safe to ignore. I'll fix them later if needed.


kiaking (approved):
I've moved comments to simply be function comment 👍  Other than that, I trust you on this 😳  Feel free to merge when you're ready!

PR #475, opened 2024-02-20
feat(avatar): add more size option

kiaking:
Add more size option to `<SAvatar>`.

- `jumbo` - `48px` big. Adding this to be more consistent with `<SButton>` sizing.
- `fill` - Custom size that fills the parent container. When we want avatar bigger than 48px, use this.

One issue with `fill` option is that it's hard to set `font-size` for name initial when the user don't have avatar image.

@brc-dd 
Do you know any tricks to like set font size dynamically based on parent container? If not, I think we can stick to `font-size: 20px` I've set. We should only use `fill` when we need big avatar and that's quite rare case so 👀

### DIFF
--- a/lib/components/SAvatar.vue
+++ b/lib/components/SAvatar.vue
@@ -1,7 +1,15 @@
 <script setup lang="ts">
 import { computed } from 'vue'
 
-export type Size = 'nano' | 'mini' | 'small' | 'medium' | 'large'
+export type Size =
+  | 'nano'
+  | 'mini'
+  | 'small'
+  | 'medium'
+  | 'large'
+  | 'jumbo'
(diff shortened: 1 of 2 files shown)


brc-dd:
container queries 😅

kiaking:
Uhhhh can we use that now? 😳

brc-dd:
Yeah similar support to `:has()`

Or just set font-size on `.SAvatar` and use `em` on `.initial`

kiaking:
OK tested several things but yeah it's too complicated... I guess we can go with fixed `24px`. Because the font shouldn't be too big, and not too small and so hard to maintain 😅

PR #481, opened 2024-02-27
chore: enable sefirot for optimizing images across domain

brc-dd:


netlify.toml
```diff
@@ -1,3 +1,5 @@
 [build.environment]
-  NODE_VERSION = "18"
-  NPM_FLAGS = "--version"
```
brc-dd, line 3: @kiaking maybe you'll need to adjust the build command in netlify dashboard to remove pnpm install from there as it's no longer needed. That --version workaround was needed 2-3 years ago when it didn't detect package manager.

PR #486, opened 2024-03-05
fix(table): use `mini` size for state cell

kiaking:
This should have been like this from beginning to align the size with pill.

Screenshot 2024-03-05 at 9 44 29

### DIFF
--- a/lib/components/STableCellState.vue
+++ b/lib/components/STableCellState.vue
@@ -11,6 +11,7 @@ defineProps<{
   <div class="STableCellState" :class="[mode ?? 'neutral']">
     <SState
       v-if="value"
+      size="mini"
       :mode="mode"
       :label="value"
     />


brc-dd:
I think without mini it was better aligned? (I mean it's height was 24px, line height of adjacent columns is also 24px, height of avatar is also 24px)

kiaking:
Hmmmm good point. In that case, we could increase Pill size to `small` 🤔 It's a subjective manner but felt like it was too big compared to text, because pill and state looks bigger due to its background color and borders.

But, good point with avatar... For example GitHub has `20px` for both pill and avatar, but having bigger avatar makes it easier to see who they are... 🤔 ...

Asana has 20px for pill and 24px for avatar.

Hmmmm... Let me try how it feels like when we have 24px pill 🏃‍♂️

kiaking:
Yeah I think 24px pill is too big for me 😭 OK, let's go with `20px` state for now. I wanna keep avatar a bit bigger since our members photos are very similar and would like to have it big as possible 👀 WDYT?

Screenshot 2024-03-05 at 11 11 26

brc-dd (approved):
Okay fine!

PR #488, opened 2024-03-05
feat(validation): add slack channel name validation

### LINKED ISSUE #478: [Validation] Add slack channel name validation rule
We use this a lot these days. Nice to have.

NozomuIkuta:
Resolves #478


lib/validation/validators/slackChannelName.ts
```diff
@@ -0,0 +1,25 @@
+import { isString } from '../../support/Utils'
+
+export const SLACK_CHANNEL_NAME_MAX_LENGTH = 80 as const
+const SLACK_CHANNEL_NAME_INVALID_SUMBOLS = /[ ,.!@#$%^&*?(){}<>=+|・'`~/\\\[\]\"]/
```
NozomuIkuta: These constants are defined based on Deus validation logic.

https://github.com/globalbrain/deus/blob/023a5d0e021e2ef83d1a211c53cab762339dab70/centre/internal/domain/model/deal/deal_entry_dd_kick_off.go#L208-L223

NozomuIkuta: I will explain this point in another place.

tests/validation/validators/slackChannelName.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    expect(slackChannelName('+')).toBe(false)
+    expect(slackChannelName('|')).toBe(false)
+    expect(slackChannelName('・')).toBe(false)
+    // eslint-disable-next-line @typescript-eslint/quotes
+    expect(slackChannelName("'")).toBe(false)
+    expect(slackChannelName('`')).toBe(false)
+    expect(slackChannelName('~')).toBe(false)
+    expect(slackChannelName('/')).toBe(false)
+    // Single back slash can't be tested because it's an escape character
+    expect(slackChannelName('\\')).toBe(false)
```
NozomuIkuta, line 37: Just in case, I am right, ain't I?

brc-dd: That comment seems off because `'\\'` is "single" back slash. (`'\\'.length === 1`)

brc-dd (approved):

lib/validation/validators/slackChannelName.ts
brc-dd: Slightly better version for validity - `/^[\p{L}\p{M}\p{N}_-]{1,80}$/u` (some of the marks (`\p{M}`) are not valid unfortunately, so it will allow some invalid characters).

lib/validation/validators/slackChannelName.ts
```diff
@@ -0,0 +1,25 @@
...
+    return false
+  }
+
+  const {
+    prefix = '',
+    suffix = ''
+  } = options
+
+  return (
+    !SLACK_CHANNEL_NAME_INVALID_SUMBOLS.exec(value)
```
brc-dd, line 22: SYMBOLS has a typo here.

tests/validation/validators/slackChannelName.spec.ts
NozomuIkuta: I meant if I write like `'\'`, then 2nd quote is escaped and the string can't be closed.

brc-dd: Yeah that won’t work 😅

NozomuIkuta: Ah, sorry I misunderstood your comment.
You are right.

lib/validation/validators/slackChannelName.ts
```diff
@@ -0,0 +1,25 @@
+import { isString } from '../../support/Utils'
+
+const SLACK_CHANNEL_NAME_MAX_LENGTH = 80 as const
+const SLACK_CHANNEL_NAME_INVALID_SUMBOLS = /[ ,.!@#$%^&*?(){}<>=+|・'`~/\\\[\]\"]/
+
+export interface SlackChannelNameOptions {
```
kiaking, line 6: I think it's better to add separate validation for these options if we want like `startsWith` and `endsWith`. These have nothing to do with slack name validation 👀 And wouldn't add so much value. It's easy as doing:

```ts
useV(data, {
  name: {
    startsWith: startsWith('x_'),
    slackChannelName: slackChannelName()
  }
})
```

lib/validation/validators/slackChannelName.ts
NozomuIkuta: Fixed: dc85c2e

lib/validation/validators/slackChannelName.ts
kiaking: Ah sorry wasn't looking at code 😓 OK I you wanted this to calculate the max length.

But in that case I see no point separating it into `prefix` and `suffix`. It could just be 1 single option 👀  Maybe `subtractLength`?

tests/validation/validators/slackChannelName.spec.ts
NozomuIkuta: Fixed: 7eb6c05

lib/validation/validators/slackChannelName.ts
NozomuIkuta: Ah, sorry, my implementation was not aligned with the options provided.
The validator should have checked `value.startsWith(options.prefix)` and `values.endsWith(options.suffix)`.

I provided the options so that we can validate Slack channel name with one rule.

Providing `startsWith` and `endsWith` rules itself would be beneficial for other cases, but providing `slackChannelName` as a wrapper of 4 rules (`startsWith`, `endsWith`, `maxLength`, (and `regExp`)) could make the usage clear? 🤔 

```ts
useV(data, {
  name: {
    startsWith: startsWith('x_'),
    slackChannelName: slackChannelName(2) // meaning of 2 looks implicit?
  }
})
```

lib/validation/validators/slackChannelName.ts
NozomuIkuta: @kiaking 

What do you think?

lib/validation/validators/slackChannelName.ts
kiaking: Hmmm I disagree with it is wrapping 4 rules. `maxLength` and `regExp` is specific to Slack. Only slack validation rule knows what those are.

But `startsWith` and `endsWith` has nothing to do with slack. If we add these, there would be nothing stopping us from adding other rules too. Why not add `required` option too? You know what I mean 😳 

And yeah I think `slackChannelName(2)` is hard to understand what this `2` is 😅  I'm not 100% happy with `subtractLength` but I think it's still better.

lib/validation/validators/slackChannelName.ts
kiaking: Sure why not! As long as test is passing. It should be fine 👀

NozomuIkuta: Refined validator based on the regular expression: 45ddb11

JFYI, here is a ref about the regular expression: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape

Note that discussion on implementation is still ongoing at https://github.com/globalbrain/sefirot/pull/488#discussion_r1513648136

@brc-dd What do you think about the conversation above?

lib/validation/validators/slackChannelName.ts
brc-dd: When do we need the prefixes? 👀  (Like does x_ has some special meaning?)

kiaking: Yeah, we have a use case where we want user to input their preferred channel name but it has be prefixed with `x_`. So, we only accept max 80 - 2 characters in this case 👀

In ordinary app, I think using computed value or `maxLength` would be more straight forward. Something like:

```ts
const input = ref('')
const prefix = 'x_'

const channelName = computed(() => {
  return `${prefix}${input.value}`
})

useV({
  input,
  channelName
}, {
  channelName: {
    slackChannelName: slackChannelName()
  }
})

// or

useV(data, {
  channelName: {
    maxLength: maxLength(SLACK_CHANNEL_NAME_MAX_LENGTH - 2)
  }
})
```

But, this is going to be tough when building form from Json schema (which we do these days

NozomuIkuta: In the latter example, it is needed to check characters available for Slack channel name (rule name doesn't matter)?

```ts
useV(data, {
  channelName: {
    slackChannelChars: slackChannelChars(),
    maxLength: maxLength(SLACK_CHANNEL_NAME_MAX_LENGTH - 2)
  }
})
```

kiaking: Yes I was just omitting it.

NozomuIkuta: I see.
I just thought it should have been there because we are talking (at least partially) about which we should choose, one wrapper rule or separated single purpose rules.

kiaking: > I just thought it should have been there because we are talking (at least partially) about which we should choose, one wrapper rule or separated single purpose rules.

Sorry if I confused you... but I don't remember talking about adding prefix or suffix rule for this validation 🤔

lib/validation/validators/slackChannelName.ts
```diff
@@ -0,0 +1,16 @@
...
+export function slackChannelName(value: unknown, options: SlackChannelNameOptions = {}): boolean {
+  if (!isString(value)) {
+    return false
+  }
+
+  const { offset = 0 } = options
+  const maxLength = /* Slack channel name max length */ 80 - offset
+
+  return new RegExp(`^\[\\p\{L\}\\p\{M\}\\p\{N\}_-\]\{1,${maxLength}\}\$`, 'u').test(value)
+}
```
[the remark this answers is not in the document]
kiaking: Yup! Looking good to me 👍

lib/validation/validators/slackChannelName.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function slackChannelName(value: unknown, options: SlackChannelNameOptions = {}): boolean {
+  if (!isString(value)) {
+    return false
+  }
+
+  const { offset = 0 } = options
+  const maxLength = /* Slack channel name max length */ 80 - offset
+
+  return new RegExp(`^\[\\p\{L\}\\p\{M\}\\p\{N\}_-\]\{1,${maxLength}\}\$`, 'u').test(value)
```
brc-dd, line 15: escaping brackets is not needed:

```suggestion
  return new RegExp(`^[\\p{L}\\p{M}\\p{N}_-]{1,${maxLength}}$`, 'u').test(value)
```

NozomuIkuta: Fixed: a23fe7e

kiaking (approved):

brc-dd (approved):

PR #497, opened 2024-03-11
feat(table): allow disabling selection by user for certain rows

### LINKED ISSUE #496: [Table] Add disable select option
I need a way to disable selecting rows on `<STable>`. Not sure how the API should be but maybe add option to `Table`?

```ts
interface Table<
  O extends string = string,
  R extends Record<string, any> = any,
  SR extends Record<string, any> = any
> {
  ...
  isRowSelectable?(record: R): boolean
  ...
}

const table = useTable({
  ...
  // For example, enable bulk selection only for rows
  // with the post that is deletable.
  isRowSelectable(post) {
    return post.isDeletable
  }
})
```

brc-dd:
closes #496

revert story before merging

### DIFF
--- a/lib/components/STable.vue
+++ b/lib/components/STable.vue
@@ -140,15 +140,13 @@ const selectedIndexes = reactive(new Set())
 
 const control = computed({
   get() {
-    const selected = indexes.value.filter((index) => {
-      return selectedIndexes.has(index)
-    })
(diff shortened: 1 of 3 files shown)


kiaking (approved):
Looking good to me! Reverted story (I kept selected ref to show checkboxes in playground 👍

PR #499, opened 2024-03-11
feat(table): highlight selected row

### LINKED ISSUE #483: [Table] Add a way to highlight selected row
When making table selectable, it would be cool if we can highlight that row. Not sure how API should look like.

I think the bg color should be set to `--c-bg-mute-1` in order to distinguish from hover style. But maybe it makes texts harder to read... so `--c-bg-elv-4` could be enough 🤔

brc-dd:
closes #483

revert story before merging


kiaking (approved):
Actually I was still switching it back and forth between elv-4 and mute-1 until now but... I think elv-4. Let's go with this!

PR #504, opened 2024-03-19
feat: unify error handling and add sentry integration

### LINKED ISSUE #476: [Error] Add unified error handling feature
Add error handling that works with Sentry. Maybe we need different module for Nuxt?

Currently, every app is doing slightly different things when catching and reporting global errors. Would be nice to have more unified way to setup.

## Idea

In Vue, error can happen on many different phases, but at the moment, we only care about error happening when running the Vue app. We can ignore error happening outside of Vue, for example css file is not loaded due to network request kinda things.

Next, inside Vue app, we usually only care catching errors inside "page". In reality, error could happen at almost anywhere. Like `app.use()` can throw error if the plugin is messed up. But those errors are very rare and usually caught during local dev, at least for now, we can ignore it.

So, ideally, I think having this kind of setup would make sense inside layout file.

```vue
<script setup lang="ts">
import { useErrorHandling } from 'sefirot/composables/Error'
import { useMe } from '@/composables/Auth'

// Get signed in user.
const me = useMe()

// Handle all app errors. Use `onErrorCaptured` inside.
// We could use `<SErrorBoundary>` as alternative.
//
// Make it able to pass in authenticated user info so that
// we can attach it when reporting to sentry.
const { error } = useErrorHandling(me)
</script>

<template>
  <div class="LayoutMain">
    <LayoutMainHeader />

    <div class="main">
      <!-- Show error page when there is a global error -->
      <ErrorPage v-if="error" :error="error" />
      <RouterView v-else />
    </div>

    <LayoutMainFooter />
  </div>
</template>
```

The `useErrorHandling` composable should catch all errors and report to Sentry. How sentry is integrated should also be thought, but this is one of `useErrorHandling` composable we have in our app.

```ts
/**
 * Capture an exception and send it to Sentry, and show error page.
 */
export function useError(): (err: Error) => void {
  return (err) => {
    config.app.env === 'local'
      ? console.error(err)
      : Sentry.captureException(err)
  }
}

export function useErrorHandling(): ErrorHandling {
  const createError = useError()

  const error = shallowRef<Error | null>(null)

  onErrorCaptured(handleError)

  // Setup `unhandledrejection` in order to catch error thrown inside of
  // promises since `onErrorCaptured` hook will not catch them.
  onMounted(() => {
    window.addEventListener('unhandledrejection', handleRejection)
  })

  onUnmounted(() => {
    window.removeEventListener('unhandledrejection', handleRejection)
  })

  function handleError(err: Error): false | void {
    error.value = err

    if (!(err instanceof PageNotFoundError)) {
      createError(err)
    }

    return false
  }

  function handleRejection(event: PromiseRejectionEvent): void {
    event.preventDefault()
    handleError(event.reason)
  }

  return {
    error
  }
}
```

brc-dd:
closes #476

# Usage

## Nuxt

```ts
// plugins/2-sentry.client.ts

import { useErrorHandler } from 'sefirot/composables/Error'
import { useAuth } from '@/composables/Auth'

export default defineNuxtPlugin({
  name: 'sentry-plugin',
  setup: (nuxtApp) => {
    const { public: { sentry } } = useRuntimeConfig()
    const { data: user } = useAuth()
    if (!sentry.dsn) { return }

    const handler = useErrorHandler({
      ...sentry,
      user: () => ({
        id: user.value.user?.id,
        username: user.value.user?.name,
        email: user.value.user?.email
      })
    })

    nuxtApp.hook('vue:error', handler)
    nuxtApp.hook('app:error', handler)
  }
})
```

```ts
// nuxt.config.ts

  runtimeConfig: {
    public: {
      sentry: {
        dsn: process.env.SENTRY_DSN,
        environment: process.env.SENTRY_ENVIRONMENT
      }
    }
  },
```

### useError / clearErrror / createError

Use Nuxt's builtins:

```ts
throw createError({ statusCode: 403, statusMessage: 'Authorization failed' })
throw createError({ statusCode: 404, statusMessage: 'Page not found' })
throw createError({ statusCode: 500, statusMessage: 'Unexpected error occurred' })
```

- https://nuxt.com/docs/api/composables/use-error
- https://nuxt.com/docs/api/utils/clear-error

## Vite

```ts
// main.ts

import { createPinia } from 'pinia'
import { useErrorHandler } from 'sefirot/composables/Error'
import { createApp } from 'vue'
import { useAuth } from '@/stores/Auth'
import App from './App.vue'

const pinia = createPinia()
const app = createApp(App)
app.use(pinia)

const userStore = useAuth() // should be called after installing pinia

app.config.errorHandler = useErrorHandler({
  dsn: import.meta.env.VITE_SENTRY_DSN,
  environment: import.meta.env.VITE_SENTRY_ENVIRONMENT,
  user: () => ({
    id: userStore.user?.id,
    username: userStore.user?.name,
    email: userStore.user?.email
  })
})

// ...

app.mount('#app')
```

### useError / clearErrror

```ts
import { useError } from 'sefirot/stores/Error'

const error = useError()

error.data

await error.clear({ redirect: '/' })
```

### createError

```ts
import { AuthorizationError, PageNotFoundError, UnexpectedError } from 'sefirot/errors'

throw new AuthorizationError()
throw new PageNotFoundError()
throw new UnexpectedError()
```


kiaking:
Since this is composable, should we place it under `sefirot/lib/composables`? Or at least, `useErrorHandler` function should be 👀 

Another thing I'm curious is that, how would we integrate this with regular error handling?

For example, let's say we want to show global error in layout. I'm currently imagining to do something like this, but maybe there is a better way.

```vue
<script setup lang="ts">
// Not sentry, generic error handler.
import { useErrorHandling } from 'sefirot/composables/Error'

const { error } = useErrorHandling()
</script>

<template>
  <div class="Layout">
    <!-- Show error page if error is captured -->
    <ErrorPage v-if="error" :error="error" />
    <RouterView v-else="" />
  </div>
</template>
```

```ts
// In `Error.ts`
function useErrorHandling() {
  const error = ref<Error | null>(null)

  onErrorCaptured((err) => {
    // Not sure how to implement this but something like this?
    if (err.statusCode === 404) {
      error.value = new ErrorType(404)
    }

    // Prevent crushing error. How do we integrate Sentry?
    // Like for some error, do not report to sentry etc. (ignore 404)
    return false
  })

  return {
    error
  }
}
```

brc-dd:
Ah, I’ll push onError code. See slack updates 😅

lib/composables/Error.ts
```diff
@@ -0,0 +1,185 @@
...
+  type ComponentPublicInstance,
+  type ConcreteComponent,
+  type MaybeRefOrGetter,
+  type VNode,
+  toValue
+} from 'vue'
+import { useError } from '../stores/Error'
+import { isFunction } from '../support/Utils'
+
+export { useError } from '../stores/Error'
```
kiaking, line 32: Should we import stores as store, or wrap with composable. I would like to discuss it a bit. I'm always not sure... but these days thinking, maybe it would be just OK to import it from where it belongs 🤔  So import from `@/stores/Error` instead of `@/composables/Error`.

What do you think? Then we need to import `useErrorHandler` and `useError` from different path, but... I don't know. But the upside is that it makes it clear that `useError` is "global" state.

kiaking (commented):
Overall, looking good to me! Can't wait to integrate this into our apps. Made a comment which is more of a convention topic 👀

lib/composables/Error.ts
brc-dd: Yeah makes sense I guess. We won't be importing both useErrorHandler and useError in same file anyway.

kiaking: Nice. OK let's do that. So, remove this re-export and I think we are good to go 🔥

kiaking (approved):
LGTM! Please merge it when ever you are ready 👍

PR #516, opened 2024-04-10
feat(action-list): support showing tooltip and implement disabled prop

### LINKED ISSUE #514: [ActionList] Add "disabled" feature
It already has `disabled` option but it does nothing yet. Let's implement it!

Apply `--c-text-3` color to both label and icon. Like this.

Screenshot 2024-04-05 at 18 18 48

---

And also add `:tooltip` prop as same as `<SButton>` so that we can display the reason why it is disabled.

```ts
interface Props {
  tooltip?: Tooltip
}

interface Tooltip {
  tag?: string
  text?: MaybeRef<string | null>
  position?: Position
  trigger?: 'hover' | 'focus' | 'both'
  timeout?: number
}
```

brc-dd:
- depends on #513
- depends on #515
- closes #514

TODO:

- [ ] revert story before merging


kiaking (commented):
Sorry I looked into this one before #515. Can we merge this once and close #515?

I've adjusted `display` of tooltip wince it wasn't being `block` and wasn't filling up entire width when tooltip is present.

Could you test it again, and if everything works great, please revert story and merge it 👍

brc-dd:
Ah, display block breaks SAvatarStack

brc-dd:
Fixed it. Check once and please revert the action list story. We can keep table one if you want 👀

kiaking (approved):

kiaking:
@brc-dd 
Looking good to me! I've adjusted the font size/weight on action list item, and reverted action list story. Let's keep table story as is 👍

PR #517, opened 2024-04-15
feat(validation): add `requiredYmdIf` and `requiredHmsIf` validators and rules

NozomuIkuta:
No related issues.

We came to dynamically validate Ymd and Hms values in our product.
This is just to add such validators and rules, `requiredYmdIf` and `requiredHmsIf`.

This PR also includes some bug fixes for the existing `requiredYmd` and `requiredHms` and some related refactoring.


lib/validation/validators/requiredHms.ts
```diff
@@ -1,9 +1,6 @@
...
+import { type Hms, HmsMap, type HmsType } from '../../support/Day'
+import { hms } from './hms'
 
 export function requiredHms(value: unknown, required: HmsType[] = ['h', 'm', 's']): boolean {
-  if (!isHms(value, required)) {
-    return false
-  }
-
-  return required.every((r) => value[HmsMap[r]] != null)
+  return hms(value, required) && required.every((r) => (value as Hms)[HmsMap[r]] != null)
```
NozomuIkuta, line 5: Oh, wait, I found that it's written in the docs that "This validator will not validate if the values are valid numbers" ([ref](https://sefirot.globalbrains.com/validation/validators#requiredymd)).

Let me revert the change.

lib/validation/validators/requiredYmd.ts
```diff
@@ -1,9 +1,6 @@
...
+import { type Ymd, YmdMap, type YmdType } from '../../support/Day'
+import { ymd } from './ymd'
 
 export function requiredYmd(value: unknown, required: YmdType[] = ['y', 'm', 'd']): boolean {
-  if (!isYmd(value, required)) {
-    return false
-  }
-
-  return required.every((r) => value[YmdMap[r]] != null)
+  return ymd(value, required) && required.every((r) => (value as Ymd)[YmdMap[r]] != null)
```
NozomuIkuta, line 5: Oh, wait, I found that it's written in the docs that "This validator will not validate if the values are valid numbers" ([ref](https://sefirot.globalbrains.com/validation/validators#requiredymd)).

Let me revert the change.

NozomuIkuta: Fixed: bb4e512

lib/validation/validators/requiredHms.ts
NozomuIkuta: Fixed: bb4e512

lib/validation/validators/requiredHmsIf.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  if (typeof condition === 'string' && condition) {
+    return requiredHms(value, required)
+  }
+
+  if (typeof condition === 'function' && (await condition())) {
+    return requiredHms(value, required)
+  }
+
+  return isHms(value, required)
```
kiaking, line 18: This should be true I guess 🤔 Otherwise, the validator might return false even if the condition is not met. When condition doesn't meet, it should not validate anything at all.

tests/validation/rules/requiredHmsIf.spec.ts
```diff
@@ -0,0 +1,92 @@
+import { requiredHmsIf } from 'sefirot/validation/rules'
+import { type RequiredIfCondition } from 'sefirot/validation/validators'
+
+describe('validation/rules/requiredHmsIf', () => {
+  function getTestCasesForAllTypes() {
+    return [
+      { value: { hour: '01', minute: '02', second: '03' }, expected: { required: true, unrequired: true } },
```
kiaking, line 7: As commented on the validator, I don't think it makes sense to have `required: false` condition.

If condition is false, then everything should return true 🤔

docs/validation/validators.md
```diff
@@ -135,7 +135,46 @@ const time = {
...
+  hour: '10',
+  minute: null, // Value missing.
+  second: '00'
+}
+
+requiredHmsIf(time, true) // <- false
+requiredHmsIf(time, false) // <- true
+```
+
+You may pass 3rd argument to specify which fields to validate. For example,if you pass `['h', 'm']`, it will validate only hour and minute.
```
cuebit, line 167:
```suggestion
You may pass 3rd argument to specify which fields to validate. For example, if you pass `['h', 'm']`, it will validate only hour and minute.
```

docs/validation/validators.md
```diff
@@ -135,7 +135,46 @@ const time = {
...
+
+```ts
+const time = {
+  hour: '10',
+  minute: '30',
+  second: null // Valud missing, but this will be ignored.
```
cuebit, line 173:
```suggestion
  second: null // Value missing, but this will be ignored.
```

docs/validation/validators.md
```diff
@@ -170,9 +209,49 @@ const date = {
...
+  year: 1985,
+  month: null, // Value missing.
+  date: 10
+}
+
+requiredYmdIf(date, true) // <- false
+requiredYmdIf(date, false) // <- true
+```
+
+You may pass 3rd argument to specify which fields to validate. For example,if you pass `['y', 'm']`, it will validate only year and month.
```
cuebit, line 241:
```suggestion
You may pass 3rd argument to specify which fields to validate. For example, if you pass `['y', 'm']`, it will validate only year and month.
```

docs/validation/validators.md
```diff
@@ -170,9 +209,49 @@ const date = {
...
+
+```ts
+const date = {
+  year: 1985,
+  month: 10,
+  date: null // Valud missing, but this will be ignored.
```
cuebit, line 247:
```suggestion
  date: null // Value missing, but this will be ignored.
```

lib/validation/validators/requiredHmsIf.ts
brc-dd: Yeah this validator will fail if value is nullish and condition is falsy. The value shouldn't be required if condition is not true / doesn't resolve to true.

tests/validation/validators/requiredHmsIf.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  function getTestCasesForGivenTypes() {
+    return [
+      { value: { hour: '23', minute: '59', second: '59' }, expected: { required: true, unrequired: true } },
+      { value: { hour: '23', minute: '59', second: '60' }, expected: { required: true, unrequired: true } },
+      { value: { hour: '23', minute: '59' }, expected: { required: true, unrequired: true } },
+      { value: { hour: '23', minute: null }, expected: { required: false, unrequired: true } },
+
+      { value: undefined, expected: { required: false, unrequired: false } },
+      { value: null, expected: { required: false, unrequired: false } },
```
brc-dd, line 30: These two should be true on unrequired probably. Same for maybe `value: {}` too. 

For `value: boolean | string | number` it can be anything. I don't have hard preference there. If we are doing validation for value every where, then probably makes sense to return false here too for consistency. Otherwise if we are ignoring different types then maybe just true.

docs/validation/validators.md
NozomuIkuta: Fixed: 46e2656

docs/validation/validators.md
NozomuIkuta: Fixed: 46e2656

docs/validation/validators.md
NozomuIkuta: Fixed: 46e2656

docs/validation/validators.md
NozomuIkuta: Fixed: 46e2656

lib/validation/validators/requiredHmsIf.ts
NozomuIkuta: Fixed: 21c1168

tests/validation/rules/requiredHmsIf.spec.ts
NozomuIkuta: Fixed: 21c1168

tests/validation/validators/requiredHmsIf.spec.ts
NozomuIkuta: Now everything is treated as valid if `condition` is true: 21c1168

brc-dd (approved):

kiaking (approved):

PR #519, opened 2024-04-22
feat(api): allow executing queries without assigning result to data

brc-dd:
Adds `assign` option to `useQuery.execute`. It's helpful while implementing functionality like "load more".

When assign is false, `loading.value` will not be set to `true` before fetching - this is to avoid showing spinner - and calling execute won't update `query.data` - one would need to manually merge the returned data with existing one.

Example:

```ts
const query = useQuery(...)

async function loadMore(): Promise<void> {
    const res = await query.execute({ assign: false })
    ;(query.data.value ??= {}).push(...res)
  }
}
```

### DIFF
--- a/docs/composables/api.md
+++ b/docs/composables/api.md
@@ -10,7 +10,18 @@ Executes given promise function immediately and set the result to `data` ref.
 interface Query<Data = any> {
   loading: Ref<boolean>
   data: Ref<Data | undefined>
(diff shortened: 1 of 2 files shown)


brc-dd:
> When assign is false, loading.value will not be set to true before fetching - this is to avoid showing spinner

This can be split to separate option like `silent` too for better customization 👀

brc-dd:
updated

PR #522, opened 2024-04-24
feat(http): support sending files along with json

brc-dd:
don't merge yet, I'll test it out first

TODO
- [ ] update docs

### DIFF
--- a/lib/http/Http.ts
+++ b/lib/http/Http.ts
@@ -15,6 +15,7 @@ export interface HttpOptions {
   xsrfUrl?: string | false
   client?: HttpClient
   lang?: Lang
+  payloadKey?: string
 }
 
 export class Http {
@@ -22,6 +23,7 @@ export class Http {
   private static xsrfUrl: string | false = '/api/csrf-cookie'
   private static client: HttpClient = ofetch
   private static lang: Lang | undefined = undefined
+  private static payloadKey = '__payload__'
 
   static config(options: HttpOptions) {
     if (options.baseUrl) {
@@ -35,6 +37,9 @@ export class Http {
     }
     if (options.lang) {
       Http.lang = options.lang
+    }
+    if (options.payloadKey) {
+      Http.payloadKey = options.payloadKey
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/http/Http.ts
```diff
@@ -117,13 +122,27 @@ export class Http {
...
   async upload<T = any>(url: string, body?: any, options?: FetchOptions): Promise<T> {
-    const formData = this.objectToFormData(body)
+    return this.post(url, this.objectToFormData(body), options)
+  }
 
-    return this.performRequest<T>(url, {
-      method: 'POST',
-      body: formData,
-      ...options
+  async postOrUpload<T = any>(url: string, body?: any, options?: FetchOptions): Promise<T> {
```
brc-dd, line 128: maybe name this something else, or merge with Http.post behind an option 👀

kiaking: Yeah I was thinking to say that too 😆 

I think it's good to have different method since it does something special.

How about... `multipartJson`...? (OK C'mon I don't like it too! It's your turn!

brc-dd: GPT ideas:

```
- postWithFileHandling
- postWithUploadSupport
- postWithMultipart
- postWithFileAwareness
- enhancedPost
- multipartPost
- fileAwarePost
- uploadPost
- formDataPost
- postWithFileSupport
- postWithFormData
- postWithUploadHandling
```

cuebit: The naming is determined by its intended use. It should coincide with how it’s interpreted in the backend because of its exclusivity. `postOrUpload` suggests one or the other, when in reality it could be both. I’m assuming this method is to accommodate files while stringifying the adjacent payload?

`postWithPayload` or `postWithAttachment` come to mind which acknowledges the distinction of the formatted payload.

The alternative is to merge with `post` itself with extended options, and make use of backend middleware which can gracefully resolve the contents from the given key.

kiaking:
Ahah. OK let me know if I'm understanding the code here.

So, when we use `post` method and pass in FormData (usually containing file), then it will create this stringified json data. And When `upload` is used, it will append non binary data just as key-value to the formdata...?

brc-dd:
`upload` is not changed. `upload` is called with body like `{ foo: File, bar: 1 }`,  then corresponding form data will have `foo = File` and `bar = '1'` and will be sent using POST request.

If `post` is called with body like `{ foo: File, bar: 1 }`, then corresponding form data will have `foo = File` and `__payload__ = '{"bar":1}'` and will be sent using POST request.

kiaking:
Gotcha! Make sense 👍

kiaking (approved):
LGTM! Please merge and release whenever you're ready.

PR #531, opened 2024-05-23
feat(input-textarea): allow enabling auto resizing of textarea on supported browsers

brc-dd:


lib/components/SInputTextarea.vue
```diff
@@ -48,6 +49,12 @@ const classes = computed(() => [
   { fill: props.rows === 'fill' }
 ])
 
+const style = computed(() => {
+  return props.autoResize
+    ? `field-sizing: content; max-height: ${props.autoResize === true ? 10 : props.autoResize}lh;`
```
brc-dd, line 54: can't use object style syntax because it's still a non-standard attribute

kiaking (commented):
@brc-dd 
I've updated the code a bit to take `rows` into account. I want it to set `min-width`, even when auto-resize is on. For example start from 1 line, etc.

I think I managed to do it, but... what do you think? Still we shouldn't use `fill` together with auto-resize.

If this looks good, I'll update the docs as well 👀

brc-dd:
Tested the story, seems to work fine.

kiaking:
Thanks! And yup I meant `min-height` 😅

kiaking:
I'll update the docs

kiaking (approved):

kiaking:
@brc-dd 
Updated! Please merge this if it looks good to you 🙏

PR #534, opened 2024-06-03
feat(indicator): add `<SIndicator>` component

kiaking:
I've changed props a bit from the issue.

```vue
<SIndicator state="running" />
```

```ts
interface Props {
  size?: Size
  state: State
  mode?: Mode
}

type Size =
  | 'nano'
  | 'mini'
  | 'small'
  | 'medium'
  | 'large'
  | 'jumbo'

type State =
  | 'pending'
  | 'ready'
  | 'queued'
  | 'running'
  | 'completed'
  | 'failed'

type Mode =
  | 'colored'
  | 'monochrome'
```

Screenshot 2024-06-03 at 17 50 03

### DIFF
--- a/lib/components/SIndicator.vue
+++ b/lib/components/SIndicator.vue
@@ -67,11 +67,11 @@ const stateIconDict = {
 .SIndicator.jumbo  { width: 48px; height: 48px; }
 
 .SIndicator.queued {
-  animation: indicator-blink 1s linear infinite;
+  animation: indicator-blink 1.5s cubic-bezier(0.45, 0.05, 0.55, 0.95) infinite;
 }
 
 .SIndicator.running {
-  animation: indicator-spin 1s linear infinite;
+  animation: indicator-spin 1.5s linear infinite;
 }
 
 .SIndicator.colored {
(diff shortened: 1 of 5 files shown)


lib/components/SIndicator.vue
```diff
@@ -0,0 +1,89 @@
...
+.SIndicator.large  { width: 40px; height: 40px; }
+.SIndicator.jumbo  { width: 48px; height: 48px; }
+
+.SIndicator.queued {
+  animation: indicator-blink 1s linear infinite;
+}
+
+.SIndicator.running {
+  animation: indicator-spin 1s linear infinite;
+}
```
brc-dd, line 75: maybe increase their duration a bit, they look too fast currently.

brc-dd (approved):
rest looks good

lib/components/SIndicator.vue
brc-dd: `1400ms cubic-bezier(0.45, 0.05, 0.55, 0.95) infinite` looks better (keep .running linear)

kiaking: Updated to 1.5s, since we use 0.25, 0.5, 0.75 scales for animations 👀  And added that bezier. Yeah it looks nicer 👍

brc-dd (approved):

PR #542, opened 2024-06-24
feat(validation): allow undersocre in domain name when validating email

kiaking:
There was a client who had email address that contains underscore in domain name. And seems like it is valid. So, updating email validation to allow unerscore.


cuebit:
AFAIK underscore does not exist in a valid FQDN. Does the underscore in the client email reside within a different label i.e. foo_bar.example.com?

kiaking:
Seems like underscore is valid as domain, and invalid as a host name. The client address was more like `john@abc_d.com`

brc-dd:
Are they even able to use their email? Because according to RFC 5321 (SMTP) it's invalid.

kiaking:
Seems like it. Our capitalist is actually communicating through that email address 🫠  So, we kinda have no choice 😆

brc-dd:
ah ok, seems like there are 6 email standards 🫠 I think we can merge this for now, or just remove this (`return value.includes('@')`) and let laravel handle it.

cuebit:
@brc-dd looks like you’ll have to fork tldts 😂

brc-dd:
I'll port https://github.com/egulias/EmailValidator to JS some day 😅

brc-dd:
Ok wait, but that package is also rejecting this 🤣 @kiaking Laravel will throw for this email.

kiaking:
Ahhh... @brc-dd Well this fix will apply to non Laravel app so it's OK for now, but damn. We might have to create custom validation for Laravel then 🫠

PR #556, opened 2024-09-25
feat(desc-text): add linkify option

### LINKED ISSUE #555: [Desc] Add option to make links clickable in `<SDescText>`
Add option to make links, email address inside `<SDescText>` to be clickable link (`<a>` tag).

Should be possible to use `<SMarkdown>` within `<SDescText>`?

```ts
// SDescText prop
interface Props {
  value?: string | null
  lineClamp?: string | number
  preWrap?: boolean
  markdown?: boolean // New! Default `false`
}
```

## Security

Is it 100% safe to parse user input as Markdown even it is inline only...? 🤔

brc-dd:
closes #555

### DIFF
--- a/docs/components/desc.md
+++ b/docs/components/desc.md
@@ -246,6 +246,20 @@ You may also use `:pre-wrap` to preserve line breaks in the text. This is especi
 </SDesc>
 ```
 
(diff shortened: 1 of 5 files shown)


kiaking:
Added story and docs 👍

kiaking (approved):

PR #560, opened 2024-10-08
feat(content): add h3 styles

brc-dd:


kiaking (approved):
I've adjusted the bullet style. Other than that, looking great 👍

PR #563, opened 2024-11-12
feat(validation): format numbers to make them more readable

### LINKED ISSUE #545: [Validation] add comma separator to length error message
So it would be easier to read. `1000` -> `1,000`.

Screenshot 2024-08-02 at 10 32 04

brc-dd:
closes #545


lib/validation/rules/maxLength.ts
```diff
@@ -2,13 +2,15 @@ import { createRule } from '../Rule'
...
 
 export const message = {
-  en: (length: number) => `The value must be less than or equal to ${length} characters.`,
-  ja: (length: number) => `この値は最大${length}文字までです。`
+  en: (length: string) => `The value must be less than or equal to ${length} characters.`,
+  ja: (length: string) => `この値は最大${length}文字までです。`
 }
 
 export function maxLength(length: number, msg?: string) {
+  const formatted = Intl.NumberFormat().format(length)
```
kiaking, line 10: Will this always result in EN locale formatting? Looking at the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat), this might change the formatting locale depending on user's browser settings?

So, maybe we should do this?

```suggestion
  const formatted = Intl.NumberFormat('en-US').format(length)
```

And if we need to specify the locale like this, maybe it would make sense to add this to `@/support/Num` package? 👀

brc-dd: Lol I just realized `support/Num` already has this 😅

kiaking (approved):

PR #565, opened 2024-11-18
feat(content): support html tables in SContent

brc-dd:


kiaking (approved):
Working smooth! Updated the docs and made tiny adjustment where aligning content to the top except for thead th.

PR #572, opened 2025-01-22
feat(table): add cell type "Path"

kiaking:
Add a cell type to display "Path" like data (directory path, breadcrumbs, etc).

```ts
const data = [
  { path: ['Folder A', 'file-a.png'] },
  { path: ['Folder B', 'file-b.png'] }
]

const table = useTable({
  records: data,
  orders: ['path'],
  columns: {
    path: {
      label: 'Path',
      cell: (_, record) => ({
        type: 'path',
        items: record.path.map((p) => ({
          label: p,
          link: `https://example.com/${p}`
        }))
      })
    }
  }
})
```

---

Screenshot 2025-01-22 at 13 00 31

### DIFF
--- a/lib/composables/Table.ts
+++ b/lib/composables/Table.ts
@@ -120,10 +120,10 @@ export interface TableCellNumber<V = any, R = any> extends TableCellBase {
 
 export interface TableCellPath extends TableCellBase {
   type: 'path'
-  items: TableCellPathItem[]
+  segments: TableCellPathSegment[]
 }
 
-export interface TableCellPathItem {
+export interface TableCellPathSegment {
   text: string
   link?: string | null
   color?: TableCellValueColor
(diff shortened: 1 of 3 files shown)


lib/composables/Table.ts
```diff
@@ -116,6 +118,18 @@ export interface TableCellNumber<V = any, R = any> extends TableCellBase {
   onClick?(value: V, record: R): void
 }
 
+export interface TableCellPath extends TableCellBase {
+  type: 'path'
+  items: TableCellPathItem[]
```
kiaking, line 123: @brc-dd 
Do you think `items` is good 🤔 Feels too bit generic I think but things like `segments` felt too long too 🫠

brc-dd: yeah segment would've been better 😅

brc-dd (dismissed):

lib/composables/Table.ts
kiaking: lol OK, I'll update this to segment. And also `TableCellPathItem` -> `TableCellPathSegment`

PR #574, opened 2025-01-27
feat(input-file-upload): add droppable and indicator support

kiaking:
Add droppable file selection and individual item state support to `<SinputFileUpload>`.

In order to keep the backward compatibility, the model value accepts both `File[]` and the new `FileObject[]`, and must define `:model-type="object"` when using `FileObject[]` as model value.

```vue
<script setup lang="ts>
const { data } = useData({
  files: [
    { file: File, indicatorType: 'running', canRemove: false }
  ]
})

const rules = {
  maxFileSize: maxFileSize('1mb')
}
</script>

<template>
  <SInputFIleUpload
    model-type="object"
    :v-model="data.files"
    :rules="rules"
  />
</template>
```

---

Screenshot 2025-01-27 at 16 46 00

### DIFF
--- a/lib/components/SInputFileUpload.vue
+++ b/lib/components/SInputFileUpload.vue
@@ -89,7 +89,7 @@ const { isOverDropZone } = useDropZone(dropZoneEl, {
 const _value = computed(() => {
   return props.modelValue !== undefined
     ? props.modelValue
-    : props.value !== undefined ? props.value : [] as unknown as ModelValue<T>[]
+    : props.value !== undefined ? props.value : [] as ModelValue<T>[]
 })
 
 const input = ref<HTMLInputElement | null>(null)
--- a/lib/components/SInputFileUpload.vue
+++ b/lib/components/SInputFileUpload.vue
@@ -1,4 +1,6 @@
-<script setup lang="ts">
+<script setup lang="ts" generic="T extends ModelType = 'file'">
+import { type ValidationRuleWithParams } from '@vuelidate/core'
+import { useDropZone } from '@vueuse/core'
 import { type Component, computed, ref } from 'vue'
 import { useTrans } from '../composables/Lang'
 import { type Validatable } from '../composables/Validation'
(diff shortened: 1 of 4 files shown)


kiaking:
@brc-dd 
Could you take a look on types especially. It is working but I'm using generics here and maybe there are room for improvements. I don't know 🤔

lib/components/SInputFileUpload.vue
```diff
@@ -121,8 +166,31 @@ function onRemove(index: number) {
...
-        <SCardBlock class="header">
+        <SCardBlock v-if="droppable" class="drop-zone" ref="dropZoneEl">
+          <div class="drop-zone-box">
+            <STrans lang="en">
+              <div class="drop-zone-text">
+                <div>Drag & Drop files here</div>
+                <div>Or</div>
+              </div>
+              <div class="drop-zone-action">
+                <SButton size="mini" label="Select files" />
```
brc-dd, line 177: missing handler

kiaking:
@brc-dd 
Yeah it happens on HMR, but can't reproduce on normal data modification. How did you do it? Does it happen outside HMR? Like adding a button to push/remove items or update indicator states works 👀 

```ts
// These works.
function addItem() {
  data.value.files.push({
    file: new File([''], `000iii111jjjjjjj.txt`, { type: 'text/plain' }),
    indicatorState: 'completed',
    canRemove: true
  })
}

function update(){
  data.value.files[1].indicatorState = null
}
```

brc-dd:
https://github.com/user-attachments/assets/a5115180-adb4-4bfb-94ee-893d3576af12

kiaking:
Ahhhh

kiaking:
Yeah this must be Histoire issue... When ever updating props, the file object becomes empty object for some reason 🤔 No idea why it happens.

```ts
// before
[File [object, object]]

// when any props gets updated
[{}]
```

stories/components/SInputFileUpload.01_Playground.story.vue
```diff
@@ -27,7 +24,8 @@ function state() {
     placeholder: 'Total 10MB max.',
```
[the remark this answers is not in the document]
kiaking: Updated! To "1MB max per file.".

lib/components/SInputFileUpload.vue
kiaking: Thanks fixed! But, the handler is added to the parent div so that the user can click the entire box 👀

brc-dd:
yeah

lib/components/SInputFileUpload.vue
```diff
@@ -50,28 +79,46 @@ const { t } = useTrans({
...
+const { isOverDropZone } = useDropZone(dropZoneEl, {
+  multiple: true,
+  onDrop: (files) => onDrop(files)
+})
+
 const _value = computed(() => {
   return props.modelValue !== undefined
     ? props.modelValue
-    : props.value !== undefined ? props.value : []
+    : props.value !== undefined ? props.value : [] as unknown as ModelValue<T>[]
```
brc-dd, line 92: as unknown assertion is probably not needed, [] should be directly castable to ModelValue<T>[]

kiaking: You are right. Fixed! I thought it was showing error but probably fixed when I changed something 🫠

brc-dd (dismissed):

PR #584, opened 2025-02-21
feat(input): add warning props to text, textarea, and number

kiaking:
I'm not proud of my self but added `:warning` props to text, textarea, and number input components. Also refactored those to use base props exported from base input (there were just so many input to do it in one shot 😅 )

Screenshot 2025-02-21 at 15 28 50

### DIFF
--- a/stories/components/SInputText.01_Playground.story.vue
+++ b/stories/components/SInputText.01_Playground.story.vue
@@ -15,7 +15,7 @@ const { data, init } = useData({
   name: null as string | null
 })
 
-const { validation, validateAndNotify } = useValidation(data, {
+const { validation, validateAndNotify, reset } = useValidation(data, {
   name: {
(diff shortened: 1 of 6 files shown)


brc-dd (approved):

stories/components/SInputText.01_Playground.story.vue
brc-dd, line 1:
```
      <HstCheckbox
        title="error"
        v-model="state.error"
      />
```

this is not used, maybe we can swap it with warning

brc-dd: other stuff is also not used, align, disabled, separator?

kiaking: Hah... yeah w might wanna organize these... 🫠

PR #586, opened 2025-02-26
fix(lang): provide lang at app level so that it's available in stores

brc-dd:


lib/composables/Lang.ts
```diff
@@ -26,7 +26,10 @@ export function useSetupLang(): (user?: HasLang | null) => void {
 }
 
 export function provideLang(lang: Lang) {
-  provide(SefirotLangKey, lang)
+  const instance = getCurrentInstance()
+  if (instance) {
+    instance.appContext.provides[SefirotLangKey] = lang
+  }
```
brc-dd, line 32: we can add some warning in else, but it always will be available, condition is just a formality

kiaking (approved):

PR #587, opened 2025-02-27
feat(http): support configuring stringify options

ryo-gk:
In some cases, we'd like to pass options to the stringify method when processing query parameters for HTTP requests.
For example, some servers accept array query parameters in the format `a=b&a=c&a=d`, while others require a different format like `a[0]=b&a[1]=c&a[2]=d`.
In such cases, we'd like to pass the `arrayFormat` option to the stringify method.

### DIFF
--- a/lib/http/Http.ts
+++ b/lib/http/Http.ts
@@ -71,8 +71,8 @@ export class Http {
     return xsrfToken
   }
 
-  private async buildRequest(url: string, fetchOptions: FetchOptions = {}): Promise<[string, FetchOptions]> {
-    const { method, params, query, ...options } = fetchOptions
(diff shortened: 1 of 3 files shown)


brc-dd (approved):

lib/http/Http.ts
brc-dd, line 1: we can also add stringifyOptions to `config` method, so that it can be configured globally per project if needed 👀

kiaking: Yeah make sense. Probably we want this 👀 

Let me add and merge it.

brc-dd (dismissed):

brc-dd (approved):

kiaking (approved):

PR #589, opened 2025-03-03
feat: add charts

brc-dd:
image

Variants:

- Bar Chart - Horizontal
   
   image

- Pie Chart
   
   image

- Donut/Pie Chart - Half
   
   image

- Donut/Pie Chart - With Labels
   
   image

- Donut/Pie Chart - With Active Slice 
   
   image

- Light mode
   
   image

\
[Story](https://deploy-preview-589--sefirot-story.netlify.app/story/stories-components-schart-01-playground-story-vue?variantId=_default)

Notes:
- svg margins have sensible defaults wherever possible, but it might need bit of hit-and-trial when using the components depending on the content
- in pie chart, enabling both legend and labels at same time is not supported and there might be some content overlap


kiaking:
@brc-dd 
Very impressive! Really cool. Can we add these options?

- Control thickness of the bar, both bar and donut.
- Add option to control font size of label and legends.
    - Including tick values in bar char like 2022, 2023, 2024, etc
    - So that we can have smaller font size when chart is small
- Maybe change `:mode` prop to `:type` or `:variant`? Because in Sefirot, mode usually refers to color style in other components 👀 I think we don't need to change `:colors` to `:mode` since `:colors` actually take multiple colors. If we were to add `:mode`, maybe it would be something like `mode="monochrome"` kinda thing 🤔 
- Maybe set Pie chart border radius to 0?

Other than this, it looks very very good. After this, let me adjust a tiny bit on stylings like tooltip padding, default thickness of the bar by talking to Keity how he wants it to be.

kiaking (approved):
OK this looks amazing! Maybe we might wanna adjust default values based on our needs, but I think we can do that once we have more use cases 👍

PR #605, opened 2025-04-22
feat(chart): auto-calculate margins in bar chart

brc-dd:


brc-dd:
One assumption, that the longest key will also occupy maximum width. This won't be true in case of multi-byte sequences and html entities.

kiaking (approved):

PR #609, opened 2025-05-01
fix(input-select): apply `:size` prop

kiaking:
`:size` was not working for `<SInputSelect>` so fixing it. Also;

- Use BaseProp from `<SInputBase>`
- Adjust "help" text padding top `6px` -> `4px` (just a styling issue I noticed).

### DIFF
--- a/lib/components/SInputSelect.vue
+++ b/lib/components/SInputSelect.vue
@@ -1,12 +1,18 @@
 <script setup lang="ts">
 import IconCaretDown from '~icons/ph/caret-down-bold'
 import IconCaretUp from '~icons/ph/caret-up-bold'
-import { type Component, computed, ref } from 'vue'
-import { type Validatable } from '../composables/Validation'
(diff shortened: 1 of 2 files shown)


brc-dd (dismissed):

lib/components/SInputSelect.vue
```diff
@@ -80,10 +69,11 @@ function emitChange(e: any): void {
...
-    :info="info"
-    :help="help"
+    :size
+    :label
+    :note
+    :info
+    :help
     :check-icon="checkIcon"
     :check-text="checkText"
     :check-color="checkColor"
```
brc-dd, line 79: not sure, but most likely these can be changed to shorthand too 😅

kiaking: Ah, good point. Gonna do it!

kiaking: Done!

PR #610, opened 2025-05-09
fix: make the `records` type inference in `useTable` call work

### LINKED ISSUE #599: [Table] Improve type inference using NoInfer where necessary

haoqunjiang:
- ref #599

I'm not sure of the exact cause of the original issue.

But my guess is that the `computed()` call complicated the type inference process significantly—especially for the `cell` field, which is the return value of callback functions that are themselves the return value of a generic function (`computed<T>`). TypeScript loses track of the connection between the several `R` generic types in the `useTable` call and, therefore, throws an error.

By using `computed()` only at the lowest necessary level, we can simplify and make the type inference process work.


kiaking:
Wow thank you so much for the PR man! 🎉

kiaking:
Actually, this does not fix the root issue because there are cases where we need to make `columns` computed 🤔 But, let's merge this for now at least it fixes the story component.

kiaking (approved):

PR #612, opened 2025-05-20
feat(validation): add `before`, `beforeOrEqual`, `after`, `afterOrEqual` rules

### LINKED ISSUE #591: [Validation] Add before/after rules (date)
Let's follow Laravel validation names 👀 

- `before`
- `beforeOrEqual`
- `after`
- `afterOrEqual`

I think for now, we should only support `Day` object and forget about `Ymd` and `Hms` type.

brc-dd:
fixes #591


lib/validation/rules/after.ts
```diff
@@ -0,0 +1,15 @@
+import { createRule } from '../Rule'
+import { after as baseAfter } from '../validators/after'
+
+export const message = {
+  en: 'The date must be after the specified date.',
```
kiaking, line 5: Can we have actual date value shown in message here? I think we can access the specified date like this 👀  We do this in things like `maxValue` validation.

```ts
export const message = {
  en: (max: string) => `The value must be less than or equal to ${max}.`,
  ja: (max: string) => `この値は最大${max}です。`
}

export function maxValue(max: number, msg?: string) {
  return createRule({
    message: ({ lang }) => msg ?? message[lang](format(max)),
    optional: true,
    validation: (value) => baseMaxValue(value, max)
  })
}
```

brc-dd: Updated. Please check the Japanese translations once.

kiaking: Thanks! I'll adjust the JA message and merge it 👍

PR #616, opened 2025-05-22
feat(pagination): add disabled prop to SPagination and SControlPagination

ryo-gk:
There are some cases where we want to explicitly disable the pagination buttons (the "prev" and "next" buttons), such as when waiting for a network request to complete.

To handle this, add a `disabled` prop to `SPagination` and `SControlPagination`.

### DIFF
--- a/lib/components/SPagination.vue
+++ b/lib/components/SPagination.vue
@@ -11,12 +11,14 @@ export type Align = 'left' | 'center' | 'right'
 
 const props = withDefaults(defineProps<{
   size?: Size
+  disabled?: boolean
   align?: Align
   total: number
   page: number
   perPage: number
 }>(), {
   size: 'medium',
+  disabled: false,
   align: 'center'
 })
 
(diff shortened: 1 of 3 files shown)


lib/components/SPagination.vue
```diff
@@ -49,11 +51,11 @@ const hasNext = computed(() => {
 })
 
 function prev() {
-  hasPrev.value && emit('prev')
+  !props.disabled && hasPrev.value && emit('prev')
 }
 
 function next() {
-  hasNext.value && emit('next')
+  !props.disabled && hasNext.value && emit('next')
```
brc-dd, line 58: this is already handled by SButton, can be removed from here?

https://github.com/globalbrain/sefirot/blob/ceb5fbfdfe53649ed659f59753937310a78ec90f/lib/components/SButton.vue#L90

lib/components/SPagination.vue
```diff
@@ -49,11 +51,11 @@ const hasNext = computed(() => {
 })
 
 function prev() {
-  hasPrev.value && emit('prev')
+  !props.disabled && hasPrev.value && emit('prev')
```
brc-dd, line 54:
```suggestion
  hasPrev.value && emit('prev')
```

brc-dd (approved):

PR #620, opened 2025-05-27
fix(input-dropdown): show the dropdown even inside container with overflow hidden

### LINKED ISSUE #618: [Flyout] Make Flyout element use position fixed
There are many cases where we want to sue flyout (dialogs) inside `overflow: hidden` element. For example, this is a horizontally scrollable tableish UI. The currency dropdown gets hidden at the bottom.

Image

For this particular issue, I want `<SInputDropdown>` to show its dropdown option using `position: fixed` so that it comes outside of the overflow hidden element.

However, I remember we had some component that does this 🤔  I'm thinking we organize that code, then make it easy to apply to any element, then apply it to `<SInputDropdown>` as well.

brc-dd:
fixes #618

TODO:
- [x] revert story

### DIFF
--- a/lib/components/SInputDropdown.vue
+++ b/lib/components/SInputDropdown.vue
@@ -181,8 +181,10 @@ function handleArray(value: OptionValue) {
         </div>
       </div>
 
-      <div v-if="isOpen" class="dropdown" :class="position">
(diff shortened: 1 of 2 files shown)


kiaking (dismissed):
Tested and working! Could you revert story and merge?

PR #637, opened 2025-07-24
feat(button,input): add new `sm` and `md` size

kiaking:
Added new sizes.

- Button: `xs` | 'sm' | 'md' | 'lg' | 'xl'
- Inputs: `sm` | 'md'

We still keep old values so no breaking changes. Adding only `sm` | `md` to input because we reraly use others.

@brc-dd 
Can you please do rough testing 🙏  I've updated relevant stories so you should be able to test in Histoire.

### DIFF
--- a/lib/components/SButton.vue
+++ b/lib/components/SButton.vue
@@ -6,7 +6,17 @@ import SLink from './SLink.vue'
 import SSpinner from './SSpinner.vue'
 import STooltip from './STooltip.vue'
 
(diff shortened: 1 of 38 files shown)


kiaking:
Added new size system support for <SControl> too.

brc-dd (approved):
LGTM

maybe update SButtonGroup too and the SButton sizes story.

kiaking:
Thanks! Right, missed that. Gonna update 🫠

PR #638, opened 2025-07-24
fix(http): config leaking state in ssr

brc-dd:
BREAKING CHANGE: `Http.config` is removed, migrate to `useSetupApp`

### DIFF
--- a/lib/composables/Api.ts
+++ b/lib/composables/Api.ts
@@ -1,3 +1,4 @@
+import { useHttpConfig } from 'sefirot/stores/HttpConfig'
 import { type Ref, type WatchSource, ref, watch } from 'vue'
 import { Http } from '../http/Http'
 import { tryOnMounted } from './Utils'
@@ -45,6 +46,7 @@ export function useQuery<Data = any>(
 ): Query<Data> {
   const loading = ref(false)
   const data = ref<Data | undefined>()
+  const httpConfig = useHttpConfig()
 
   if (options.immediate !== false) {
     tryOnMounted(execute)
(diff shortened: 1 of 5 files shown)


kiaking:
Thanks! Though, I think it would be better for Http class to not depend on store, since store is composable, this will make Http class not usable outside of Vue components (outside of Vue App context).

Well we don't have a such usecase 😅 but, I think the general idea is that anything under composables/stores directory with `useXXX` prefix should not be called outside of Vue component. Which means if it is not prefixed by `useXXX`, it should be safe to call it from anywhere (like in top of `main.ts`).

So, I believe approach something like this would be more straight forward as a composable usage?

```ts
// Make `Http` class pure. Do not depend on Store.
export class Http {
  private config: HttpOptions

  constructor(options: HttpOptions) {
    this.config = options
  }
}

// In Api composables, create fresh instance by passing options from the stores.
export function useQuery() {
  const httpConfig = useHttpConfig()
  // ...

  async function execute({ assign = true, silent = false } = {}): Promise<Data> {
    // ...

    const res: Data = await req(new Http(httpConfig))
    if (assign) { data.value = res }

    // ...
  }

  // ...
}

// For global setup, change `useSetupHttp()`
// composable in `/composables/Http.ts`
export function useSetupHttp(): (user?: HasLang | null, options?: HttpOptions) => void {
  const browserLang = useBrowserLang()
  const httpConfig = useHttpConfig()

  return (user, options = {}) => {
    httpConfig.apply({
      lang: user?.lang ?? browserLang,
      ...options
    })
  }
}
```

This way, we can safely use Http in anyplace by `new Http({ ... }).get(...)`.

This will be breaking change, but for apps only that are not using `useSetupHttp()` at the moment so should be quick fix 👀 

What do you think?

kiaking (approved):

PR #639, opened 2025-08-01
feat(input): add `tabindex` props to basic input components

### LINKED ISSUE #528: Add `tabindex` props to basic input components
In some cases, we would like to set the "tabindex" to some components.
So how about adding `tabindex` props on some Input compoents?
I guess it's enough the following components have it.

- SInputText
- SInputTextarea
- SInputNumber
- SInputSelect
- SInputDate
- SInputFile

haoqunjiang:
Closes #528

### DIFF
--- a/lib/components/SInputDate.vue
+++ b/lib/components/SInputDate.vue
@@ -20,6 +20,7 @@ const props = defineProps<{
   checkColor?: Color
   block?: boolean
   disabled?: boolean
+  tabindex?: -1 | 0 | number
   modelValue: Day | null
   validation?: Validatable
   hideError?: boolean
@@ -86,6 +87,7 @@ function emitBlur() {
           autocomplete="off"
           :value="inputValue"
           :disabled
+          :tabindex
           v-on="disabled ? {} : inputEvents"
           @blur="emitBlur"
         >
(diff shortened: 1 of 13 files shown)


stories/components/SInputText.01_Playground.story.vue
```diff
@@ -164,12 +169,16 @@ function initState() {
           :placeholder="state.placeholder"
           :unit-before="state.unitBefore"
           :unit-after="state.unitAfter"
+          :align="state.align"
+          :disabled="state.disabled"
```
haoqunjiang, line 173: Took me quite a while to realize that these other props aren't added to the component, despite the controls being there… Though it was a bug or something 😅
Need to check if other playgrounds have the same issue too.
But for now, I think it suffices to only add the missing ones when we encounter any during development.

kiaking: Ah, thanks!

lib/components/SInputDate.vue
```diff
@@ -20,6 +20,7 @@ const props = defineProps<{
   checkColor?: Color
   block?: boolean
   disabled?: boolean
+  tabindex?: -1 | 0 | number
```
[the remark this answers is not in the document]
kiaking: 👍

kiaking (approved):

PR #640, opened 2025-08-01
refactor(input): re-export Size & Color types from SInputBase

### LINKED ISSUE #623: consistent Size and Color type exports across all components
- ref https://github.com/globalbrain/fatima/pull/314#discussion_r2120133880

We have started to extends base props from `<SInputBase>` on many input components, however not all of it. Let's organize these props and make it clean.

One discussion needed. We had comment from @brc-dd on whether we should re-export some props like `Size` from inherited components. What should we do?

> Size and Color type exports are inconsistent between input components. It used to be there in SInputDropdown, but in that size prop PR we removed it. Probably should be re-exported across all components instead of re-declaring in each. Needs a fix in vue-tsc, though, re-exports don't work with generic components.

haoqunjiang:
closes #623

I also updated the `CheckColor` type to be a re-export of `Color` as the values always match.

`SInputImage`, `SInputFileUpload`, and other non-input components have different sizes than `SInputBase` so they are kept as-is.

There's a `Color` type export in `SInputSegments` that has an additional `default` value. But I don't see this type used anywhere after searching the codebase, so I guess it's redundant and I removed the type completely.


brc-dd (approved):

brc-dd:
I think CheckColor can be renamed to just Color. It aligns better with other components - https://github.com/search?q=repo:globalbrain/sefirot+/checkColor%5C?:+Color/&type=code (vs https://github.com/search?q=repo:globalbrain/sefirot+/checkColor%5C?:+CheckColor/&type=code)

Fine to be a breaking change. Don't think any app is using that prop anyway.

lib/components/SInputImage.vue
```diff
@@ -4,10 +4,10 @@ import { type Component, computed, ref } from 'vue'
 import { useImageSrcFromFile } from '../composables/Image'
 import { type Validatable } from '../composables/Validation'
 import SButton from './SButton.vue'
-import SInputBase from './SInputBase.vue'
+import SInputBase, { type Color as CheckColor } from './SInputBase.vue'
 
 export type Size = 'mini' | 'small' | 'medium'
-export type CheckColor = 'neutral' | 'mute' | 'info' | 'success' | 'warning' | 'danger'
+export type { CheckColor }
```
brc-dd, line 10: this left

haoqunjiang: Oops. Fixed. Thanks!

PR #643, opened 2025-08-08
fix(action-menu): flip dropdown alignment if not enough space 

### LINKED ISSUE #632: [ActionMenu] Position dropdown dialog based on position
Currently, `<SActionMenu>` dialog is positioned left, so if we put action menu button on right side of the screen, it gets cut off. Would be great if it gets positioned automatically.

Image

haoqunjiang:
Fixes #632


brc-dd:
It should prefer the dropdown-align prop if there is space:

image

PR #471, opened 2024-02-14
feat(input-textarea): add preview functionality

brc-dd:
https://deploy-preview-471--sefirot-story.netlify.app/story/stories-components-sinputtextarea-01-playground-story-vue?variantId=_default

We can probably leave `.prose ...` styles to be implemented at consumer-side because they largely will be common with the actual post rendering.

PR #655, opened 2025-11-10
feat(SInputTextarea): add an "actions" slot to display action bar

haoqunjiang:
From this comment: https://github.com/globalbrain/sefirot/pull/471#issuecomment-1951900164

Couldn't come up with a good UX to add a YouTube video without using an action bar, so I ended up implementing this in Sefirot.

Preview at: https://deploy-preview-655--sefirot-story.netlify.app/story/stories-components-sinputtextarea-03-actions-story-vue?variantId=_default

### DIFF
--- a/lib/components/SInputTextarea.vue
+++ b/lib/components/SInputTextarea.vue
@@ -1,8 +1,5 @@
 <script setup lang="ts">
 import { computed, ref } from 'vue'
-
-import SControl from './SControl.vue'
-import SControlRight from './SControlRight.vue'
 import SInputBase, { type Props as BaseProps } from './SInputBase.vue'
 import SInputSegments from './SInputSegments.vue'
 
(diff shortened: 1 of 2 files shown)


brc-dd (dismissed):

kiaking:
Also, it would be awesome if we can make the right margin (padding) consistent with the Y spacings. But I guess that should be controlled by the users...? 🤔  In that case, maybe it might make sense to not wrap the content in `SControl` by the component. But maybe we could just simply define it as slot with display flex so that it doesn't get fixed padding on X axis.

Screenshot 2025-11-11 at 9 22 08

```html
<div v-if="preview !== undefined || $slots.actions" class="control">
  <SInputSegments
    v-if="preview !== undefined"
    :options="[
      { label: writeLabel ?? 'Write', value: false },
      { label: previewLabel ?? 'Preview', value: true }
    ]"
    v-model="isPreview"
    size="mini"
  />

  <SControl v-if="$slots.actions && !isPreview" class="actions">
    <SControlRight>
      <slot name="actions" />
    </SControlRight>
  </SControl>
</div>

<style scoped lang="postcss">
.actions {
  display: flex;
  align-items: center;
  gap: 8px;
}
</style>
```

haoqunjiang:
image
image

It feels more visually balanced with a 8px padding though.

PR #656, opened 2025-11-17
feat(STable): better loading animation with skeleton screen

haoqunjiang:
Preview: https://deploy-preview-656--sefirot-story.netlify.app/story/stories-components-stable-03-loading-story-vue?variantId=_default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Changelog
 
+## [4.34.1](https://github.com/globalbrain/sefirot/compare/v4.34.0...v4.34.1) (2026-01-19)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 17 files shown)


brc-dd:
maybe reset head's scroll too, if the table is scrolled horizontally and the page is changed it gets misaligned:

image

haoqunjiang:
🤔 2 options to fix this issue:
- Patch the vue package for now; it was a very insignificant bug fix that just happens to trigger a bug (or design flaw?) in histoire. Patching it should be fine. It's harder to fix the issue on the histoire side.
- Moving the core logic of the Story to another non-story component so that histoire won't try to assign the `computed` fields. This is a cleaner fix, but it messes up our folder layout a bit.

brc-dd:
Monkey patched `@vue/reactivity` on main for now. 🫠 We can remove it when we migrate from histoire.

PR #674, opened 2026-01-29
fix(STable): account for scrollbar width in table header padding when body is scrollable

haoqunjiang:
To test this on macOS, go to Settings -> Appearance, set "Show scrollbars" to "Always".

https://deploy-preview-674--sefirot-story.netlify.app/story/stories-components-stable-03-loading-story-vue?variantId=_default

### DIFF
--- a/stories/components/STableLoadingExample.vue
+++ b/stories/components/STableLoadingExample.vue
@@ -503,8 +503,16 @@ function updateTagsFilter(value: string) {
 }
 
 .table :deep(.container.body) {
-  /* keep the table body inside the viewport to be scrollable */
-  max-height: calc(100vh - 56px * 2 - 41px - 32px * 2 - 56px);
+  /* Keep the table body inside the viewport to be scrollable */
+  max-height: calc(100vh -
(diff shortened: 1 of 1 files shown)


stories/components/STableLoadingExample.vue
```diff
@@ -525,7 +529,7 @@ function updateTagsFilter(value: string) {
 }
 
 .table :deep(.col-actions) {
-  --table-col-position: sticky;
+  /* --table-col-position: sticky; */
```
kiaking: Should we remove this instead of commenting out?

kiaking: Ah, good point.

kiaking: @haoqunjiang 
Can you revert the story? And then feel free to merge 👍

haoqunjiang: I commented out the sticky part to mimic the real-world use case, but forgot to un-comment it 😅

Done. And added comments about the height calculation.

PR #679, opened 2026-02-18
chore: update vite to 7, vitest to 4, and histoire to 1.0 beta

haoqunjiang:
It's mostly type fixes because histoire 1.0 is more strict on types.


kiaking:
@brc-dd 
Can you also review just in case. If it looks good to you, please feel free to merge 👍

PR #732, opened 2026-06-09
feat(lens): include the active language in catalog search requests

kiaking:
## Summary

`LensCatalog` now includes the active display language in the payload it sends with every search request — `settings.lang`, read once from `useLang()`. This lets a backend localize values it has to resolve on its own (for example, the title shown for a related-record column) to the language the user is currently viewing.

`settings.lang` is already part of the `LensQuery` type (`LensQuerySettings`), so this only populates an existing field — there are no type or schema changes, and backends that ignore it are unaffected.

## Notable changes

- The language is read once at setup via `const lang = useLang()` and passed into the request input as `settings: { lang }`, rather than read inline.

## Test plan

- [ ] In a non-default locale, verify that a backend which localizes server-resolved strings (e.g. related-field titles) receives the language and renders them in the active locale.


kiaking:
@brc-dd @haoqunjiang 
Review ready 👍

brc-dd (approved):

